Issue #8, opened 2016-09-16
Kotlin-Reflect dependency


yoavst:
Is there a way to drop the kotlin reflection dependency? a 5 MB library for getting the name of the class from companion sounds too much :)

oshai:
any idea or direction how to do that?

oshai:
one way I think of is checking the name if it ends with "$Companion" and remove it. what do you think?

yoavst:
Since you can name companion this is not 100% working solution, but I believe it is acceptable limitation.

oshai:
A solution for that is to look for the last $ sign. do you think it is better?

oshai:
The removal of kotlin-reflect dependency will be in 1.3.3

PR #9, opened 2016-09-17
fix for #8

oshai:
removed kotlin reflection dependency
changed companion object name resolving to look explicitly in string for
the word $Companion


oshai:
found another solution see #10

PR #14, opened 2017-01-10
make parameters in LocationAwareKLogger nullable, because the slf4j-i…

ChristophGr:
…nterface allows it

to mitigrate exceptions like

Exception in thread "main" java.lang.IllegalArgumentException: Parameter specified as non-null is null: method mu.internal.LocationAwareKLogger.debug, parameter arg2
	at mu.internal.LocationAwareKLogger.debug(LocationAwareKLogger.kt)


oshai:
Thanks!

oshai:
the fix was released in 1.4.2

PR #19, opened 2017-03-15
upgrade to kotlin 1.1.0

morj:


build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.0.6'
```
oshai: is there any comptability issue for applications compiling with 1.0? I got a note in slack that it should be compiled with a flag (which I haven't found)

src/test/kotlin/mu/internal/KLoggerNameResolverTest.kt
```diff
@@ -21,7 +21,7 @@ class KLoggerNameResolverTest {
         assertEquals("mu.internal.BaseClass\$InnerClass\$Obj", KLoggerNameResolver.name(BaseClass.InnerClass.Obj.javaClass))
         assertEquals("mu.internal.BaseClass\$InnerClass", KLoggerNameResolver.name(BaseClass.InnerClass.CmpObj::class.java))
         assertEquals("mu.internal.BaseClass\$InnerClass", KLoggerNameResolver.name(BaseClass.InnerClass.CmpObj::class.java))
-        assertEquals("mu.internal.Foo\$Bar", KLoggerNameResolver.name(Foo.Bar::class.java))
```
oshai: was that test failing before or something was changed?

morj: Prior to kotlin 1.1 this Foo.Bar was an object reference. But in kotlin 1.1 it's a variable reference (because you have an object and a variable named the same). I think it's either the test semantics or the variable name must be changed.

morj:
@oshai I fixed the test properly just by removing the `val Bar = this` hack.

build.gradle
morj: I've set `apiVersion` and `languageVersion` to `1.0`.

morj: Reference: http://kotlinlang.org/docs/reference/using-gradle.html#compiler-options

oshai:
@morj - please note the test KLoggerNameResolverTest fails now.

build.gradle
```diff
@@ -57,6 +57,15 @@ bintray {
         }
     }
 }
+
+// kotlin compiler compatibility options
```
oshai: Thanks for that!

morj:
@oshai yes, sorry, I missunderstood the test semantics (remember that @JvmField trick doesn't work in 1.1), must be fixed now.

oshai:
Thanks!

PR #22, opened 2017-06-18
Suppress exceptions from toString methods

imanushin:
Sometimes classes have incorrect toString implementation, which raises exception.

To avoid this, logger should cautch and inform developer.

Such behavior is used in the many popular loggers, we should reuse it too

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ gradle-app.setting
 .idea/vcs.xml
 .idea/jsLibraryMappings.xml
 local.properties
+kotlin-logging.iml
 
 # Sensitive or high-churn files:
 .idea/dataSources.ids
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ gradle-app.setting
 .idea/dictionaries
 .idea/vcs.xml
 .idea/jsLibraryMappings.xml
+local.properties
 
 # Sensitive or high-churn files:
 .idea/dataSources.ids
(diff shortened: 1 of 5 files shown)


oshai:
Can you give an example of popular loggers frameworks that catch the exception? it seems like developers will lose information this way. Also, why not catch it at the user (of this lib) code?

imanushin:
@oshai , code which does formatting (```invoke``` function in our case) should catch exception.

Examples: 
* [NLog](https://github.com/NLog/NLog/blob/master/src/NLog/LogEventInfo.cs#L517)
* From [log4j 2](https://logging.apache.org/log4j/2.x/source-repository.html) (see FormattedMessage.getMessage): 
```
        try {
            final MessageFormat format = new MessageFormat(msgPattern);
            final Format[] formats = format.getFormats();
            if (formats != null && formats.length > 0) {
                return new MessageFormatMessage(locale, msgPattern, args);
            }
        } catch (final Exception ignored) {
            // Obviously, the message is not a proper pattern for MessageFormat.
        }
        try {
            if (MSG_PATTERN.matcher(msgPattern).find()) {
                return new StringFormattedMessage(locale, msgPattern, args);
            }
        } catch (final Exception ignored) {
            // Also not properly formatted.
        }
```


Ideally: developer should not get any exceptions from logger. There are several sets of errors:
* Formatting errors - should be caught by us, because we do the formatting
* Stream Writing errors - is done by downstream logger
* Global: OOM, StackOverflow, etc. - should be skipped

oshai:
ok, thanks. I will merge with some fixes to message itself.

PR #23, opened 2017-06-19
Update versions

imanushin:
Just versions update (and minor gradle build denormalization)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,7 +10,8 @@ buildscript {
 }
 
 plugins {
-    id "com.jfrog.bintray" version "1.7"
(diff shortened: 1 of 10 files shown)


build.gradle
```diff
@@ -51,8 +51,8 @@ bintray {
         licenses = ['Apache-2.0']
         vcsUrl = 'https://github.com/MicroUtils/kotlin-logging'
         version {
-            name = "1.4.5"
-            desc = "kotlin-logging 1.4.5 - Lightweight logging framework for Kotlin"
+            name = version
```
oshai: no idea why, but this doesnt work

PR #25, opened 2017-08-29
Update README.md example with similar syntax

krokofant:
From the description
*Call log methods, without checking whether the respective log level is enabled*
and
*Behind the scenes the expensive message do not get evaluated if debug is not enabled*
one might think that the first example which has another syntax `{}` does not perform a check at all. With the same syntax it's less confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@ A convenient and performant logging library wrapping [slf4j](http://www.slf4j.or
 
 * Call log methods, without checking whether the respective log level is enabled:
 ```Kotlin
-logger.debug { "Some $expensive message!" }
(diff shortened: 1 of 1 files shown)


oshai:
I think you missed something here or did not understand the message correct:
The call is performed like that `logger.debug { "Some $expensive message!" }`
Behind the scenes, there is a check if debug level is enabled and only then the string is evaluated. if you change it to regular brackets it will evaluate the string immediately, so it is on purpose.
Hope it is more clear, if not let me know I will try to explain again. Will love to know how the readme can be changed to explain that.

krokofant:
Oh, I see, that makes sense! 😅 Thank you for clarification.

PR #40, opened 2018-03-24
Added methods to KLogger for return value after logging

agorshkov23:


oshai:
Can you please explain what is the use case?

agorshkov23:
For example, a logging if null values in the arguments

```kotlin
fun sum(a: Int?, b: Int?): Int {
    a ?: return logger.warn({ "a must be not null" }, -1)
    b ?: return logger.warn({ "b must be not null" }, -1)
    return a + b
}
```

oshai:
I believe that this is not the responsibility of a logging framework. 
Maybe another framework for conditionals like orElse etc'?
And also it is pretty easy to add it as an extensions methods to the framework in case you disagree.

agorshkov23:
For some reason, I did not consider this functionality from the point of view of logging, but I agree with you.
And yes, the extension of the methods here is more suitable.

PR #56, opened 2018-11-21
Use Kotlin version 1.3.10

morj:


oshai:
@morj - do you need a new version with this change or can it wait for other changes?

morj:
@oshai we can surely wait, thanks for the quick response.

PR #63, opened 2019-02-13
Javascript - Extracting Message Formatter interface

robertfmurdock:
So mostly this is freehand refactoring (it'd be nice to add a test suite to the JS side here). But additionally, there are two new JS configurable variables at the root, such that someone using the library can modify or replace the message formatting.

I quick put this together because I was adding some logging to an app I've been converting to Kotlin-Js, and couldn't quite get it to produce JSON log lines rather than traditional ones. 🤷🏻‍♂️


oshai:
Thanks for your contribution!

oshai:
I added it in 1.6.24.

PR #66, opened 2019-02-15
created a KotlinLoggingConfiguration object with all configurable par…

oshai:
…ameters


oshai:
@robertfmurdock what do you think about this change?

PR #72, opened 2019-03-30
added support for same functionality as slf4j-ext.

corneil:
Turns out to get location-aware work as expected I didn't use XLogger. 
Replicated the functionality. I'm not sure if essentially copying the code is legal.

### DIFF
--- a/kotlin-logging-common/src/main/kotlin/mu/KXLogger.kt
+++ b/kotlin-logging-common/src/main/kotlin/mu/KXLogger.kt
@@ -1,0 +1,11 @@
+package mu
+
+
+expect interface KXLogger : KLogger {
+
+    fun entry(vararg argArray: Any)
+    fun exit()
+    fun <T> exit(retval: T): T where T : Any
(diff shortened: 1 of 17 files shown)


oshai:
If in any case, we're implementing the xlogger methods by hand, I don't really understand why to create another factory/class/ etc' and not just add new methods to the existing KLogger. I also don't understand why slf4j didn't take that approach. maybe a compatibility issue?
In addition, if there is no js implementation at the moment, maybe we should keep it on the jvm level only?

corneil:
After all this I'm thinking the best may be to just add the `entry`, `exit`, `throwing` and `catching` methods to the existing KLogger. Their presence doesn't add any overhead. We don't need any extra dependencies.

corneil:
As to the JS implementation, I'm not sure how much work is involved in logging the method name for entry / exit

oshai:
Sounds good!

oshai:
Its ok to start with jvm-only version.

corneil:
You can reject/close this PR.

PR #73, opened 2019-03-31
added support for same functionality as slf4j-ext.

corneil:
This PR only adds the entry, exit, throwing and catching methods to the existing logger.
The location-aware logger will add method names etc.
The LocationIgnorant and JS implementations are similar.
We may want to add markers.


oshai:
Thanks for the PR!

oshai:
released in 1.6.26

Issue #77, opened 2019-07-05
simplify the way how to define logger


LiYing2010:
Currently, we need the following code to get a instance of logger:

`private val logger = KotlinLogging.logger {} `

the code itself is very simple, but the problem is, we have to put this line in **every class** which use log, and for a large project, there could be many many classes.

So I suggest we can add a new extension property, which can get the logger instance for us, and we can use this property in every classes without calling the logger initialization code.

I made a PR #78 for this, please take a look

oshai:
I have a couple of concerns regarding the suggestion:
- The biggest one is that add a property on all classes clutter your env (ie: it will suggest it for autocompletion in any case you included the library). I really don't like that.
- probably it is better not to have synchrinze block at all (but create duplicated instances).
- I don't think it will work on the file level - only inside classes?

The first one is my main concern preventing me from merging such a PR.

LiYing2010:
>  The biggest one is that add a property on all classes clutter your env (ie: it will suggest it for autocompletion in any case you included the library). I really don't like that for libraries, in your applicative code, it is more acceptable.


no, it will not be added to all classes.
it's an extension property, it will be available only if you import it, like `import mu.log`,
without this explicit import statement, this property will not be added to class.

LiYing2010:
>  probably it is better not to have synchronized block at all (but create duplicated instances).

OK
I think I can remove synchronized block

LiYing2010:
>  I don't think it will work on the file level - only inside classes?

You are right, an extension property only works for classes,
I will do more research, to see if there is anything we can do for the `file level` case

oshai:
Regarding the import, I know it will not added to all classes unless you import. The issue is the IDE will suggest to add it and import on almost any auto-completion (because it's a public extension to all instances)

oshai:
I decided not to merge such a PR at the moment. Thanks anyway!

PR #78, opened 2019-07-05
add extension function and extension property for getting logger

LiYing2010:
this PR fix issue #77

change list:
add extension function and extension property for getting logger, to simplify the way how to define logger

with the help of this new extension property `T.log`, we can use `log` in any class, without calling the logger initialization code.

Before:
we need this logger initialization code:
`private val logger = KotlinLogging.logger {}`
before we can output log message

After this change, we can call
```
log.debug { "some message" }
```

without the logger initialization code.

for a large project, there could be many many classes, so this new extension property can save much of the duplicated code

### DIFF
--- a/kotlin-logging-jvm/src/main/kotlin/mu/KotlinLoggingExtension.kt
+++ b/kotlin-logging-jvm/src/main/kotlin/mu/KotlinLoggingExtension.kt
@@ -1,0 +1,38 @@
+package mu
+
+import mu.internal.KLoggerNameResolver
+import kotlin.reflect.KProperty
+
+inline fun <reified T : Any> T.logger(): KLogger =
+    KotlinLogging.logger(
+        @Suppress("NON_PUBLIC_CALL_FROM_PUBLIC_INLINE")
+        KLoggerNameResolver.name(forClass = T::class.java)
+    )
+
+val <T : Any> T.log: KLogger by LoggerLoader()
+
+private class LoggerLoader<T : Any> {
+    private val loggers: MutableMap<Class<*>, KLogger> = mutableMapOf()
+
+    operator fun getValue(thisRef: T, property: KProperty<*>): KLogger {
+        val instanceClass = thisRef::class.java
+        val v1 = loggers[instanceClass]
+        if (v1 !== null) {
+            return v1
+        }
+
+        return synchronized(this) {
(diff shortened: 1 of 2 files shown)


sghpjuikit:
Hello, I'm passing by and see the PR and wanted to chip in my opinion. While I see the good intention of the author of this PR, I am very against merging this into the library.

- This makes the library inconsistent in how it wants users to use logger by providing two separate mechanisms.
- This provides extension on Any, which pollutes API and should only be provided if merits vastly overshadow demerits. Not the case here, I'm afraid
- The design decision to go with extension (that is 'implicitly' invoked on `this`) is bad for the same reason scala implicit conversions are (not always) bad - everything is everything and nothing is clear. This basically implicitly (with just import) turns every type into some kind of Loggable. It is debatable whether this is clean or clean-er as other approaches, as with proper mixins (like in scala) the method has its elegance. However for logging this is, clear case of no. In fact it is very bad and prone to cause unexpected behavior. See below.
- The design decision to abandon companions for convenience is a trap. Logging is cross cutting concern and logger belongs to the class instance, not the object. Companion is the correct place to define logger in and provide it for instances. The provided extension is a sugar to call the 'companion' logger without the companion (presumably by tapping right into the global logger map) - like Java member method that secretly calls static method instead). Ignoring the discussion if this is ever desirable (convenience that abandons correctness, due to added cognitive load for the user), this is inherently dangerous:

1 The receiver the log/logger() is called on is often not clear and sometimes will result in wrong one. Consider using it in nested typesafe builder and other constructs that change semantics of this within scope. Which this do you want to be used to log? Expecting developer to pay attention to these cases in irresponsible and opens then up to nasty problems that could be avoided easily. How? By specifying the exact class the used logger is bound to by binding it to the companion object!

2 How do you expect this to work in subclasses where user would declare another logger for companion. In this case your approach using `this` instead of `this.super` or something like that would result in wrong logger instance to be used again. 

TLDR: the logger should be context bound, always, at creation time, explicitly. It gives clear semantics and expected behaviot with 0 pitfalls.

How do I know all this from top of my head? I myself used this very approach in the past, having a static method (Java) I could call like this: `log(this).error(...)`.

oshai:
Will not merge, see issue #77

PR #86, opened 2019-08-07
added separate logger.exitOpt for optional return types.

corneil:
Added separate function `exitOpt` for when the return type is optional. It is a mess otherwhise.


corneil:
Updated.

src/jsMain/kotlin/mu/KLogger.kt
```diff
@@ -1,132 +1,130 @@
...
-  /**
-   * Add a log message indicating an exception will be thrown along with the stack trace.
-   */
-  actual fun <T> throwing(throwable: T): T where T : Throwable
-
-  /**
-   * Add a log message indicating an exception is caught along with the stack trace.
-   */
-  actual fun <T> catching(throwable: T) where T : Throwable
+	/**
```
oshai: can you please fix indentation? it makes the diff big and hard to find real differences. also we don't use that indentation in other files.

oshai:
Merged with #87

PR #89, opened 2019-08-14
Provide for nullable in entry and rename exit argument

corneil:
The type for entry is `vararg Any` and should be `vararg Any?` 
The name for exit was `retval` and `result` will  be better.


oshai:
Thanks for the PR!

oshai:
released in 1.7.5

PR #110, opened 2020-03-25
Upgrade to Kotlin 1.4-M1 test

rnentjes:
This is a test/example to see how easy it is to create a library with the new IR format introduced in Koltin 1.4-M1.

This works for my scenario, some remarks:

- I needed to upgrade kotlin version to 1.2, compiler no longer supports 1.1
- Used the default naming of targets to simplify testing, backwards compatible naming probably still works fine

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,6 @@ fabric.properties
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 
+gradle.properties
+gradle.properties.tmp
+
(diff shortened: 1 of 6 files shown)


oshai:
Thanks for the PR, I will wait for the official release to merge 1.4 changes, but thanks for the heads up and effort.

oshai:
@rnentjes please let me know if you want to update the change and merge it, now that 1.4 is out.

rnentjes:
Created a new pull request with the changes for 1.4.0.

muzuro:
Any updates? I think using kotlin-logging force gradle to use 1.3.50 instead of 1.4

oshai:
see #127

Issue #45, opened 2018-08-16
Kotlin/Native Support


joelhandwell:
As issue https://github.com/MicroUtils/kotlin-logging/issues/21 was only dedicated for Kotlin/JVM and Kotlin/JS multi platform support, this issue is for Kotlin/Native support.

Kotlin/Native target platform is seen in ```org.jetbrains.kotlin.konan.target.KonanTarget.kt``` file 
[line 43](https://github.com/JetBrains/kotlin-native/blob/master/shared/src/main/kotlin/org/jetbrains/kotlin/konan/target/KonanTarget.kt#L43)  as of 2018 August 16th:

```kotlin
sealed class KonanTarget(override val name: String, val family: Family, val architecture: Architecture) : Named {
    object ANDROID_ARM32 :  KonanTarget( "android_arm32",   Family.ANDROID, Architecture.ARM32)
    object ANDROID_ARM64 :  KonanTarget( "android_arm64",   Family.ANDROID, Architecture.ARM64)
    object IOS_ARM32 :      KonanTarget( "ios_arm32",       Family.IOS,     Architecture.ARM32)
    object IOS_ARM64 :      KonanTarget( "ios_arm64",       Family.IOS,     Architecture.ARM64)
    object IOS_X64 :        KonanTarget( "ios_x64",         Family.IOS,     Architecture.X64)
    object LINUX_X64 :      KonanTarget( "linux_x64",       Family.LINUX,   Architecture.X64)
    object MINGW_X64 :      KonanTarget( "mingw_x64",       Family.MINGW,   Architecture.X64)
    object MACOS_X64 :      KonanTarget( "macos_x64",       Family.OSX,     Architecture.X64)
    object LINUX_ARM32_HFP :KonanTarget( "linux_arm32_hfp", Family.LINUX,   Architecture.ARM32)
    object LINUX_MIPS32 :   KonanTarget( "linux_mips32",    Family.LINUX,   Architecture.MIPS32)
    object LINUX_MIPSEL32 : KonanTarget( "linux_mipsel32",  Family.LINUX,   Architecture.MIPSEL32)
    object WASM32 :         KonanTarget( "wasm32",          Family.WASM,    Architecture.WASM32)

    // Tunable targets
    class ZEPHYR(val subName: String, val genericName: String = "zephyr") : KonanTarget("${genericName}_$subName", Family.ZEPHYR, Architecture.ARM32)

    override fun toString() = name
}
```

[...]

PR #119, opened 2020-06-21
Add Linux X64 Support

napperley:
This PR partly addresses issue #45 by only adding support for a single Kotlin Native target (**linuxX64**), and is the successor to PR #103 . Implementation of this PR is based on the existing JS implementation with some minor adaptations made to make it work with the Kotlin Native **linuxX64**  target. Use of dependencies is minimal with only the Kotlin Native Standard library, and POSIX being used. Most Linux distributions have built-in support for POSIX (Android is one of the main exceptions).

The latest stable version of Kotlin is used due to Kotlin Native currently not having proper backwards compatibility. It remains unknown if this situation will be resolved with the Kotlin 1.4 release that is [arriving soon](https://kotlinlang.slack.com/archives/C3SGXARS6/p1592375878474700?thread_ts=1592272044.448700&cid=C3SGXARS6).


build.gradle.kts
```diff
@@ -135,15 +148,16 @@ publishing {
                 url.set("http://github.com/MicroUtils/kotlin-logging/tree/master")
             }
         }
-        artifact(tasks["javadocJar"])
+        // TODO: Fix generation of Java Doc Jars.
```
oshai, line 151: This should be fixed before merge

src/linuxX64Main/kotlin/mu/ConsoleOutputAppender.kt
```diff
@@ -0,0 +1,15 @@
+package mu
+
+import platform.posix.fprintf
+import platform.posix.stderr
+
+object ConsoleOutputAppender : Appender {
+    override fun trace(message: Any?) = println(message)
+    override fun debug(message: Any?) = println(message)
+    override fun info(message: Any?) = println(message)
+    override fun warn(message: Any?) = println(message)
```
oshai, line 10: isn't warn should also be printed to stderr?

src/linuxX64Main/kotlin/mu/Appender.kt
```diff
@@ -0,0 +1,9 @@
+package mu
+
+interface Appender {
```
oshai, line 3: I am ok with duplication (of js code) if we don't have alternative, but isn't it possible with multiplatform to have a module which is shared by js and native? it looks like it will save a lot of copy/paste here.

src/linuxX64Main/kotlin/mu/ConsoleOutputAppender.kt
napperley: The stderr text stream is used for error messages only, whereas the stdout text stream is used for ordinary text messages (including warnings).

src/linuxX64Main/kotlin/mu/Appender.kt
napperley: Presumably you are referring to [MPP Hierarchical Project Structure](https://blog.jetbrains.com/kotlin/2020/06/kotlin-1-4-m2-released/#hierarchical-project-structure), which isn't available in Kotlin 1.3. @oshai - Doing code sharing in the **common** module could be an option, **if** you are fine with some changes being made to the **jvm** module to match the changes made to the **common** module.

build.gradle.kts
napperley: It is important to note that there are many errors that appear after running the **dokka** Gradle task.

napperley: Below are some of the elements that dokka cannot find:

- Class:KMarkerFactory in
- Class:KotlinLogging in
- Interface:KLogger in
- Function:getMarker in
- Field:INSTANCE in
- Function:logger in
- Function:trace in
- Function:debug in
- Function:info in
- Function:warn in
- Function:error in
- Function:entry in
- Function:exit in
- Function:throwing in
- Function:catching in

napperley: Looks as though there are many parts of the public API that haven't been documented to Dokka's satisfaction.

napperley: Can apply the fix in the build file however the build itself will fail CI tests.

src/linuxX64Main/kotlin/mu/Appender.kt
napperley: The **Appender** interface is now defined in the **common** module.

oshai: since as you say the hierarchial project structure is only from 1.4, I suggest to leave all duplications, and we might fix it later instead of moving files to the common module. I don't want to clutter the jvm module.

build.gradle.kts
oshai: ok, you're welcome to open another issue for the dokka errors. the main reason we need (other than documentation itself) it is that maven central don't allow to publish without docs jars).

src/linuxX64Main/kotlin/mu/ConsoleOutputAppender.kt
oshai: I am ok with your decision, but usually warn are not regular messages: https://stackoverflow.com/a/4919110/411965

src/linuxX64Main/kotlin/mu/KotlinLoggingConfiguration.kt
```diff
@@ -0,0 +1,8 @@
+package mu
+
+object KotlinLoggingConfiguration {
+    var LOG_LEVEL = KotlinLoggingLevel.INFO
```
arkivanov, line 4: `object` are automatically frozen by default, so `var`s do not make sense here. You can specify backing fields with `AtomicReference` and use getters and setters here.

src/linuxX64Main/kotlin/mu/Appender.kt
Dominaezzz: You can currently share a source set between JS and Native. The only caveat would be that the IDE would only do intelli-sense for pure Kotlin, which is acceptable in this case.

build.gradle.kts
```diff
@@ -143,7 +156,7 @@ bintray {
     user = "oshai"//project.hasProperty("bintrayUser") ? project.property("bintrayUser") : System.getenv("BINTRAY_USER")
     key = "mykey" //https://bintray.com/profile/edit
     // project.hasProperty("bintrayApiKey") ? project.property("bintrayApiKey") : System.getenv("BINTRAY_API_KEY")
-    setPublications("metadata", "jvm", "js")
+    setPublications("metadata", "jvm", "js", "linuxX64")
```
Dominaezzz, line 159: Might not immediately work with native targets. https://github.com/bintray/gradle-bintray-plugin/issues/229 .

napperley: Removed **linuxX64** from the list. There will be an opportunity to revert back after the issue is fixed with the Bintray Gradle plugin.

src/linuxX64Main/kotlin/mu/KotlinLoggingConfiguration.kt
napperley: All properties in `KotlinLoggingConfiguration` now use `AtomicReference`, and have been turned into read only properties.

src/linuxX64Main/kotlin/mu/Appender.kt
napperley: Reverted changes to Appender interface.

oshai:
@napperley is the PR ready to be merged from your perspective? If so I will create an artifact from this to verify it's ok and after we test that I will merge.

napperley:
The PR is ready to be merged.

src/linuxX64Main/kotlin/mu/KotlinLoggingConfiguration.kt
```diff
@@ -0,0 +1,9 @@
+package mu
+
+import kotlin.native.concurrent.AtomicReference
+
+object KotlinLoggingConfiguration {
+    val logLevel: AtomicReference<KotlinLoggingLevel> = AtomicReference(KotlinLoggingLevel.INFO)
```
arkivanov, line 6: Perhaps it would be better to hide atomics as implementation details and expose plain properties with getters and setters:
```
private val _prop = AtomicReference<Type>...
val prop: Type
    get() = _prop.value
    set(value) { return _prop.value = value }
```

napperley: Although this would be the way to go there are two issues. 

First issue is that Kotlin 1.4 is due to be released soon (likely in a matter of weeks). and it remains unknown if the public/private properties [feature](https://youtrack.jetbrains.com/issue/KT-14663) will be part of the Kotlin 1.4 release. If the feature is part of the release then that would make the approach a wasted effort.

Second issue is that there is some urgency to allow enough time for @oshai to test the changes before the PR is merged in a timely manner. The comment (review) appeared after a agreement was made for the PR to be merged, **after** the PR has passed a final round of tests.

arkivanov: As far as I understand this is going to be public API. If so, you won't be able to change it later without breaking binary and backward compatibility. If you make proper API now then later you can just get rid of private "backing fields" here, but the API won't change. But up to you guys, I'm just a random reviewer here.

oshai: @napperley - you're effort making this PR is well appreciated! I do see a point in what @arkivanov said and I wonder what do you think in term of the API itself?
In regards to that I am going to release a snapshot version with current implementation anyway before merging, and we can decide that after the merge and before the next release to create another PR to change it, thus not hurting backward compatibility. WDYT?

oshai:
@napperley I tried to deploy 1.8.0.1.
bintray publish seems fine but I couldn't publish to github packages with the following error:
```
Caused by: org.gradle.internal.resource.transport.http.HttpErrorStatusCodeException: Could not PUT 'https://maven.pkg.github.com/MicroUtils/kotlin-logging/io/github/microutils/kotlin-logging-js/1.8.0.1/kotlin-logging-js-1.8.0.1.jar'. Received status code 422 from server: Unprocessable Entity
```
Did you encounter something similar?

napperley:
I have encountered the **422** error before but haven't found a way to resolve it. One theory for the error is that the same artifact (GitHub package) exists on GitHub Packages therefore the publishing fails. Unfortunately GitHub doesn't provide a way to remove a individual artifact (GitHub package) from GitHub Packages.

src/linuxX64Main/kotlin/mu/KotlinLoggingConfiguration.kt
napperley: Considering the importance of API backward compatibility the extra time needed to incorporate the changes won't be a issue (this addresses the first issue). As for the second issue it might be better to do the changes anyway, and take chances with the Kotlin 1.4 release (the risk would likely be low).

napperley: All the changes have been made.

napperley:
Did a quick visual inspection of the GitHub package. Everything looks ok :+1:  .

napperley:
All is not well. Tried to use the GitHub **linuxx64** Package in a test project, and Gradle cannot find it. Using the following in the build file:

```kotlin
// ...
repositories {
    jcenter()
    mavenCentral()
    maven {
        url = uri("https://maven.pkg.github.com/MicroUtils/kotlin-logging")
    }
}

kotlin {
    linuxX64 {
        compilations.getByName("main") {
            dependencies {
                val kotlinLoggingVer = "1.8.0.3"
                implementation("kotlin-logging:kotlin-logging:$kotlinLoggingVer")
            }
        }
        // ...
    }
}
// ...
```

oshai:
I think that there is a line that should be changed:
```
implementation("kotlin-logging:kotlin-logging-linuxx64:$kotlinLoggingVer")
```

napperley:
That doesn't make a difference. Usually with Kotlin Native libraries the name of the module is omitted when adding a library as a dependency, eg:
`implementation("kotlin-logging:kotlin-logging:1.8.0.3")`

oshai:
Might be, but as far as I understand that is how gradle find the artifact. See here: https://github.com/MicroUtils/kotlin-logging/packages/286126

oshai:
Can you try also to see if there is a difference from the version you managed to publish?

napperley:
Turns out one has to [authenticate](https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-gradle-for-use-with-github-packages#installing-a-package) in order to access the GitHub Packages Maven repository which isn't going to work :frowning: .

oshai:
> Turns out one has to [authenticate](https://help.github.com/en/packages/using-github-packages-with-your-projects-ecosystem/configuring-gradle-for-use-with-github-packages#installing-a-package) in order to access the GitHub Packages Maven repository which isn't going to work 😦 .

This is strange, because via the ui you can download it without even being signed-in.

oshai:
@napperley it seems we are a bit stuck here. What do you think about merging the code changes without the changes in the build? so we actually won't publish those artifacts yet but we'll have the code ready until we solve it, maybe with changes to multiplatform framework in kotlin.

napperley:
@oshai - If you are referring to removing the configuration for GitHub Packages then I can have it removed from the build file. Hopefully artifacts hosted by GitHub Packages in the future will be publicly accessible through Gradle.

oshai:
Thanks @napperley , I merged it. Hope the packaging issue will be solved soon.

PR #125, opened 2020-08-18
Make macos logger as copy of linux logger

alekseytomin:
If you use kotlin 1.4 - you can merge macos and linux (and minGW - I'm doing it later)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -68,6 +68,7 @@ kotlin {
         }
     }
     linuxX64("linuxX64")
+    macosX64("macosX64")
     sourceSets {
         commonMain {
             dependencies {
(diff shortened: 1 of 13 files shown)


oshai:
Thanks for the PR!
I am not sure I understood your comment. If we upgrade to Kotlin 1.4 this code is mostly redundant?
If so, I think it's better to upgrade to 1.4 first and then add only the required parts.
In addition, I think we didn't manage to publish linux artefacts yet, in case you have insights about that.

alekseytomin:
On 1.4 You can merge linux/macos/mingw in the one source - https://kotlinlang.org/docs/reference/whatsnew14.html#sharing-code-in-several-targets-with-the-hierarchical-project-structure

oshai:
ok, then see #127. we will merge it before this PR.

alekseytomin:
Replaced by https://github.com/MicroUtils/kotlin-logging/pull/131

PR #141, opened 2020-09-13
Write log into file

alekseytomin:


oshai:
Hi @alekseytomin thanks for the PR!
Can you please give more details on the change, and what is exactly the use case?
My main concern is that developing a robust logging takes more than just writing to a file, and should include for example rolling appender etc'.
I am thinking out loud here if the kotlin-logging should develop those frameworks for each platform or use existing libraries for the platform. In case of jvm it was just a facade. For JS and Native we added the console appender which is pretty simple and I think that maybe if we go to more complex solutions it is better to reuse existing libs.
WDYT?

oshai:
stale

Issue #144, opened 2020-09-14
pom.xml has no dependencies anymore?


reconman:
In version 1.8.3, the pom.xml still had slf4j-api, and some kotlin libraries as dependencies, but now the pom.xml has 0 dependencies: https://repo1.maven.org/maven2/io/github/microutils/kotlin-logging/1.11.3/kotlin-logging-1.11.3.pom

What happened to cause this? Can you add the slf4j-api dependency again?

oshai:
Thanks for reporting the issue! I will check it.
There was a change in the packaging since 1.8.3 and this is probably related. It wasn't intended.

oshai:
It is related to the change done in `1.11.1` by #137 . The last version with correct pom was [1.11.0](https://repo1.maven.org/maven2/io/github/microutils/kotlin-logging/1.11.0/kotlin-logging-1.11.0.pom).

Since we have one more [issue](https://github.com/MicroUtils/kotlin-logging/issues/126#issuecomment-691916274) with 1.4, I think I will revert to Kotlin 1.3 in a branch and publish it.

oshai:
It turns out this was caused by a bug in gradle related gradle metadata feature. After Upgrading gradle to 6.6.1 this was solved.
Fixed in kotlin-logging 1.11.5.

Issue #130, opened 2020-08-28
fail to deploy after upgrade to Kotlin 1.4.0


oshai:
Issue #127 upgraded the lib to Kotlin 1.4.0.
It seems that after the push deployment is not working anymore.
I created an issue and there are more details on [youtrack](https://youtrack.jetbrains.com/issue/KT-41483).

What's important for the lib now is that all `1.9.x` and `1.10.x` versions are malformed and shouldn't be use.
The known good version is `1.8.3`, so please use this one until solved or reverted.

altavir:
The bintray plugin is known to be bugged. It is missing gradle metadata module. The module must be added manually. I've stopped using it and now I am using regular maven-publish plugin: https://github.com/mipt-npm/scientifik-gradle-tools/blob/master/src/main/kotlin/ru/mipt/npm/gradle/KSciencePublishPlugin.kt. It requires manually creating packages on the bintray for new artifacts, but otherwise, it is much more reliable.

Also since 1.4, the package naming patter is changed, `-common` should not be there anymore. You should use no suffixes for common module `-jvm` for jvm and `-js` for js. But those are managed automatically by kotlin plugin.

altavir:
Since I need the library ASAP to continue development, I've made a fork and published artifacts to the GitHub packages: https://github.com/altavir/kotlin-logging/packages.  The workaround for bintray plugin bug could be found here: https://github.com/mipt-npm/scientifik-gradle-tools/blob/6464d9d926043b4e8ed5af0c4ef27c64f0934f59/build.gradle.kts#L135

oshai:
I'll do the package rename only if it will be very complicated to keep the old schema, because this is a breaking change.
Thanks for the workaround, I will try it out.

altavir:
It is not a package name, it is an artifact name. It does not affect anything but Gradle dependency suffix.And those are dropped anyway. Since 1.4 suffixless dependencies are used both for mpp common and platform versions.

oshai:
Changing the artifact name is a breaking change and can create a mess. Imagine that you moved to the new name but a lib you depend on use the old name or vice versa. Gradle is not aware it's the same package which can cause collision of classes and issues on every tiny change due to wrong class being loaded.

oshai:
found the doc that might help: https://kotlinlang.org/docs/reference/migrating-multiplatform-project-to-14.html#for-library-authors

oshai:
This issue is fixed, although there has been some other issue #144 .

Issue #126, opened 2020-08-18
IR backend


robertfmurdock:
1.4.0 is out! Time to compile this in BOTH mode.

oshai:
Hi @robertfmurdock can you please share more details on what exactly do you mean?

robertfmurdock:
Oh sure. Kotlin 1.4 adds a new compiler backend called IR, and to support the new backend libs can be built in “both” mode to allow users to compile in whatever mode they choose. More info is available on the kotlin 1.4 update page

robertfmurdock:
I should be more specific - I’m talking about the JS IR. The JVM IR has no ‘both’ mode AFAIK, and thus isn’t great for libs as of now.

altavir:
I've currently hit the same problem. We are migrating libraries to Kotlin 1.4 and it is impossible to use kotlin-logging because it does not export IR artifacts. In order to make it work, you should update kotlin to 1.4 and select `BOTH` or `IR` option. I do not plan to support legacy mode, so `IR` mode only will be OK for me.

robertfmurdock:
@oshai Looks like things are merged... but a new release has yet to be published! 😨

oshai:
Yes, I encountered some issues while publishing. See #130 for status and details.

robertfmurdock:
Ah, makes sense. I just looked and discovered I put in the bintray workaround in my own projects last year (to support the new Gradle metadata regarding Multiplatform projects), so I never saw this problem. 😅

oshai:
If you can point to a similar example where it works it will be great.

robertfmurdock:
https://github.com/robertfmurdock/testmints/blob/2b45f7c52ecb9b530467ca0eadced338a61795c4/build.gradle.kts#L109 <https://github.com/robertfmurdock/testmints/blob/2b45f7c52ecb9b530467ca0eadced338a61795c4/build.gradle.kts#L109>

This is the relevant section in my own libs. The migration docs for library authors include a specific workaround:

https://github.com/bintray/gradle-bintray-plugin/issues/229#issuecomment-473123891 <https://github.com/bintray/gradle-bintray-plugin/issues/229#issuecomment-473123891>

Hope that helps

nunocastromartins:
I don't know what's missing, but I still can't use `kotlin-logging` v1.11.0 when compiling with IR.

I'll investigate and report here, but maybe this issue should be reopened as I don't think the PR fixed this.

[...]

PR #145, opened 2020-09-15
Build refactor and bintray plugin replacement

altavir:
Fix fo #126 #130 #144 and partially #45


nunocastromartins:
Also out of curiosity, why did you undo your `Generify native build` commit?

altavir:
The current upstream build already contains a different way to generify the build. So I decided to keep the upstream version.

nunocastromartins:
Fair enough, I'm just more used to the "default" (as in, the one that comes with a newly generated MPP lib) one.

oshai:
Thanks for the PR. As for the deployment, I guess snapshot releases are the same, how do I perform regular releases? manually?

altavir:
I do regular releases manually from gradle, but you can trigger the deploy process from CI. Exactly the same way, you were doing for bintray plugin. The only difference is that you have to use `publishAllPublicationsToBintrayRepository` instead of `bintrayUpload`. The only difference is that maven-publish does not create bintray package automatically, you need to create it manually before the first time you want to push a new artifact. You have it already so no actions needed.

altavir:
> Fair enough, I'm just more used to the "default" (as in, the one that comes with a newly generated MPP lib) one.

The difference is in naming and tests. I am not sure that native tests are running correctly in this variant. On the bright side, source set naming is the same for different OS.

nunocastromartins:
@altavir, don't know if you saw what I wrote above, what's all the `mipt-npm` stuff there?

altavir:
> @altavir, don't know if you saw what I wrote above, what's all the `mipt-npm` stuff there?

Oops. Forgot to clear it out. I will fix it right away. Those are parts of our deploy plugin, I used at first.

nunocastromartins:
Yeah, I thought so, that's why I was asking just to make sure. :)

nunocastromartins:
> don't know if you saw what I wrote above

My bad, didn't realise I needed to press "start review" for my comment to be submitted so I didn't actually "write" anything above. :')

nunocastromartins:
@oshai, if you make a snapshot version for testing let me know and I'll give it a go.

altavir:
@nunocastromartins  you can use this one: https://bintray.com/mipt-npm/dev/kotlin-logging/1.11.5-npm. It is exactly the same, only with a different repository name.

nunocastromartins:
Ah, thanks, it works fine! I mean, I was already using another version of your fork for testing IR stuff, so I had no doubts that yours would work.

My offer was mostly to check that @oshai got the publishing to work correctly on his end.

oshai:
I will create a snapshot to test ASAP.

oshai:
The snapshot is here: http://oss.jfrog.org/webapp/builds/kotlin-logging/553
Please test it.

nunocastromartins:
Everything seems to be working. :)

Just one question, are you planning to release this as a minor version or version 2.0? I noticed that you named it 1.11.5-SNAPSHOT.

oshai:
It will be version 2.0, thanks for pointing out!
I hope I will get the new deployment correct.

oshai:
@altavir - Thank you very much for the PR!

altavir:
You are welcome. Sadly, I will have to use my fork for now since I work on Windows and I need both mingw and Linux native targets (windows generates them both).

oshai:
Snapshot for 2.0.0 is here: http://oss.jfrog.org/webapp/builds/kotlin-logging/555

oshai:
@altavir I wonder if I can build windows version using travis: https://docs.travis-ci.com/user/reference/windows/
Do you know what is the proper way to deploy from multiple builds machines?

altavir:
@oshai I think that running deploy script from multiple virtual machines with different OSes should do the trick. The build script deploys artifact relevant to the system you are currently using. On windows (at least with WSL installed) it is possible to generate artifacts both for windows and for Linux, so I am deploying everything from the local machine right now. I will play around with multi-system deployment as soon as we finish adding native targets to our libraries.

oshai:
Created #149 we will see how it goes.

nunocastromartins:
Btw @oshai, forgot to report, build `2.0.0-SNAPSHOT` works well on my end.

oshai:
@altavir trying to publish I get the following error:
```
FAILURE: Build failed with an exception.

* What went wrong:
Task 'publishAllPublicationsToBintrayRepository' not found in root project 'kotlin-logging'.
```
Any idea what is missing?

altavir:
@oshai I think that the problem is with project properties. [This line](https://github.com/MicroUtils/kotlin-logging/blob/40425d25e278003a23af7fd078458b41c98781d3/build.gradle.kts#L133) ensures that bintray repository publication is not registered unless `bintrayRepo`, `bintrayUser` and `bintrayApiKey` are all defined in the properties. It seems like `bintrayOrg` is needed as well. Probably one of those are missing. You probable could hardcode `bintrayRepo` and `bintrayOrg` since it is always the same for you, but user and API key should be defined externally.

oshai:
Thanks, that seems to work!

oshai:
I released 2.0.2 to bintray and maven central. Will update docs soon.

oshai:
@altavir do you know if maven central has similar api to the bintray one so the publish plugin can be used for that as well?

altavir:
@oshai Yes, it should work exactly the same way for maven central, just with a different repository definition. Though maven central has more strict rules about artifact preparation and I never tried that. There are a lot of examples though. Like [the official one](https://docs.gradle.org/current/userguide/publishing_maven.html#publishing_maven:complete_example).

PR #148, opened 2020-09-16
upgrade kotlin to 1.4

oshai:
Apply commit c722ca8a to 1.x branch
without breaking compatibility.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,7 +2,7 @@ import org.jetbrains.dokka.gradle.DokkaTask
 import java.util.*
 
 plugins {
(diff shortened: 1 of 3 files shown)


breun:
@oshai According to https://github.com/MicroUtils/kotlin-logging#version-2x only 2.x requires Kotlin 1.4+. However, I see that this change landed in kotlin-logging 1.12.0. Is 1.12.0 still compatible with Kotlin 1.3.x?

oshai:
The 1.x branch (include 1.12.0) compiled with kotlin 1.4 but supports and compatible with kotlin >= 1.2.

breun:
Great, thanks!

PR #149, opened 2020-09-16
Add support for windows and multiple os

oshai:
will build on windows,linux and mac.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,15 @@ jdk:
 
 dist: trusty
 
+os:
+- linux
+- osx
+- windows
+
 stages:
 - name: build
 - name: snapshot
-  if: branch IN (master, pre-snapshot, 1.x-snapshot) AND type = push
+  if: branch IN (master, pre-snapshot, 1.x-snapshot, windows) AND type = push
 - name: release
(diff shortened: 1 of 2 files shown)


oshai:
It looks like Java is not supported on windows in Travis, so another alternative is to try circleci.

oshai:
circle ci only support java 11 out of the box, so need to explicitly compile to java 8:
https://stackoverflow.com/questions/49509330/kotlin-compile-to-java-1-8

petertrr:
Hey @oshai , have you considered using github actions for releases? You can use linux, macos and windows runners in the same workflow, and Java versions can be configured using `actions/setup-java`.
I have set up releasing to Maven Central using github actions recently, if you are interested you can check out my config: [PublishingConfiguration.kt](https://github.com/petertrr/kotlin-multiplatform-diff/blob/main/buildSrc/src/main/kotlin/io/github/petertrr/PublishingConfiguration.kt), [release.yml](https://github.com/petertrr/kotlin-multiplatform-diff/blob/main/.github/workflows/release.yml).

oshai:
implemented in another cl in `2.1.15`

PR #155, opened 2020-10-01
Add `import mu.KotlinLogging;` statement to getting started in README

mfickett:
Setup with this library was super easy except that it wasn't obvious from the main README what the import line for KotlinLogging should be. Adding it in the getting started example would have been handy for me, maybe it'll help others?


oshai:
Thanks!

PR #157, opened 2020-10-14
Declaring example dependencies as implementation

alllex:
Declaring dependencies as `compile` is no longer a best practice in Gradle. [See here](https://docs.gradle.org/current/userguide/java_library_plugin.html).

Instead, they should be declared as `implementation`.


oshai:
Thanks!

Issue #172, opened 2021-03-23
Support optional/nullable values for MDC


fdw:
I'd like  to add a value to the `MDC` *if it exists*. With the current implementation, I have to manually check this before I can call `withLoggingContext`. However, I'd like to pass a nullable argument that is only added to `MDC` if it's not null.

If you're interested, I would also try to write a PR.

oshai:
what kind of api did you have in mind?

fdw:
My suggestion would be a very simple `fun withLoggingContext(vararg Pair<String, String?>...)` with a `filter { it.value != null}` or something similar.

fdw:
It was so easy that I went ahead and opened a PR. Hope you don't mind :)

oshai:
Released in 2.0.7. Thanks again!

fdw:
Quick question: I can't find a 2.0.7 release. Did I misunderstand you?

oshai:
I think it's because jcenter is not working anymore because the build was working: https://travis-ci.org/github/MicroUtils/kotlin-logging/jobs/761498773

I'll have to move to use maven central directly and only then publish again.

fdw:
Ah, thank you and good luck! 🙂

PR #173, opened 2021-04-01
Support nullables in withLoggingContext

fdw:
`withLoggingContext` now accepts nullable values that won't be written
to the `MDC` (and thus also don't overwrite existing values).
This might be useful if some values are not guaranteed to exist, but you
want them in the `MDC` if possible.

Issue: #172


oshai:
Thanks for the PR! and sorry for the delay in response.
LGTM.

PR #174, opened 2021-04-09
Allow Any in withLoggingContext

fdw:
I hope it's okay that I opened a PR without an issue first 🙂

To make calls even more comfortable, `withLogginContext` now accepts `Any?` as a value and will convert it to a `String` itself.

### DIFF
--- a/src/jvmMain/kotlin/mu/KotlinLoggingMDC.kt
+++ b/src/jvmMain/kotlin/mu/KotlinLoggingMDC.kt
@@ -10,9 +10,9 @@ import org.slf4j.MDC
  * }
  * ```
  */
-public inline fun <T> withLoggingContext(pair: Pair<String, String?>, body: () -> T): T =
+public inline fun <T> withLoggingContext(pair: Pair<String, Any?>, body: () -> T): T =
     if (pair.second != null) {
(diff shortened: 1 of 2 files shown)


oshai:
I think it's better to be explicit and say we take only strings and let the user do the conversion in this case.
MDC api is also like that (not accepting Objects): http://www.slf4j.org/api/org/slf4j/MDC.html

fdw:
Sure :)

I had a situation where I had to do an explicit `toString` call, which I wanted to avoid, as it worsens the readability. And since in Kotlin, `toString` is called automatically for object in templated strings, I thought it was a nice addition.

But if you don't like it, feel free to close the PR.

BorzdeG:
You can make a separate extension module that will accept Any. I also thought to add this + wanted to add support for Enum
If you will allow me - will I expand PR?

fdw:
I'm fine with you extending the PR, but I cannot speak for @oshai :)

oshai:
> I think it's better to be explicit and say we take only strings and let the user do the conversion in this case.
> MDC api is also like that (not accepting Objects): http://www.slf4j.org/api/org/slf4j/MDC.html

@BorzdeG as mentioned here, I prefer to keep the simple api.

PR #187, opened 2021-08-14
Fixed losing previous values in MDC

BorzdeG:
- Fixed losing previous values in MDC
- Update dependencies
- Update Gradle wrapper
- Added Detekt


oshai:
Thanks for the PR!

PR #191, opened 2021-11-04
added the license to all jar files

bespaltovyj:
Issue https://github.com/MicroUtils/kotlin-logging/issues/188

Comments for solutions:

1. I removed all `apply` because plugins are already applied in `plugins` sections
2. There are 2 Jar tasks in the gradle: 
   - `org.gradle.jvm.tasks.Jar`
   - `org.gradle.api.tasks.bundling.Jar` is subtype of `org.gradle.jvm.tasks.Jar`.
In build.gradle.kts by default using `org.gradle.api.tasks.bundling.Jar`, but Kotlin multiplatform register `sourcesJar` task with type `org.gradle.jvm.tasks.Jar`. 

So I added `import org.gradle.jvm.tasks.Jar` to add a license to all jar tasks.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 import io.gitlab.arturbosch.detekt.Detekt
+import org.gradle.jvm.tasks.Jar
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
 
 plugins {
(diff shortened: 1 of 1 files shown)


oshai:
Thanks for the PR!

PR #192, opened 2021-11-18
Fix Travis CI badge in README

SebastianAigner:
This changes the URL to point to the correct Travis CI badge.


oshai:
thanks for the fix!

PR #197, opened 2021-11-19
make travis jobs run on the same stage

oshai:
to make them run in parallel


oshai:
moving from travis to gh actions

Issue #179, opened 2021-06-01
Suggestion to use reified genertic fun instead of `KotlinLogging.logger`


kosiakk:
Option to use normal function call
```kotlin
val logger = KotlinLogging.logger()
```
instead of current
```kotlin
val logger = KotlinLogging.logger {}
```
or both (maybe with deprecation and automatic migration, if community agrees)

Definition of such function is a little bit unusual because it's defined as an extention to any class:

```kotlin
inline fun <reified T> T.getLogger(): KLogger {
    val clazz = T::class.java
    val name = clazz.canonicalName ?: clazz.toString()
    return KotlinLogging.logger(name) // can be simplified with access to KLoggerFactory
}
```

I think this is better from any point of view: 
- usual syntax
- no guessing the class name by lambda class name
- the best possible performance (a bit important because logger is usually not static in Kotlin)

kosiakk:
Since `T` is a `reified` type parameter, it's known at runtime.  Also the function is `inline`, so that generated byte-code has a simple class literal. This is a decompiled version of generated code:
```
      Class clazz$iv = GlobalSchedule.class;
      String var10001 = clazz$iv.getCanonicalName();
      if (var10001 == null) {
         var10001 = clazz$iv.toString();
         Intrinsics.checkNotNullExpressionValue(var10001, "clazz.toString()");
      }

      String name$iv = var10001;
      this.logger = KotlinLogging.INSTANCE.logger(name$iv);
```

oshai:
Something is not 100% clear to me:
The signature is `KotlinLogging.logger()` although the method called `getLogger`?
In addition - isn't `T` resolved to `KotlinLogging` at runtime?

kosiakk:
1. Yes, sorry `getLogger` is just how I called it locally - I've wrote this function for me, as external to Kotlin-logging. Internally you could choose different name and pass `Class` instance directly to SLF4J (they will deal with names)

2. Due to Java generic type erasure, it's impossible at runtime dymanically. So it's statically compiled to a specific class by means of Kotlin's [reified generic types](https://kotlinlang.org/docs/inline-functions.html#reified-type-parameters). Therefore `T` is not resolved. It's a sepecific constant for each method call, baked-in by the compiler at call-site.

oshai:
ok, thanks. I suggest to create a PR to add such functionality.
We can discuss later if we want to deprecate other apis.

ArcticLampyrid:
[...]

PR #205, opened 2021-12-09
Added capability to use a reified inline function for instantiating a logger

severn-everett:
Addressing issue #179

### DIFF
--- a/src/jvmTest/kotlin/mu/KotlinLoggingTest.kt
+++ b/src/jvmTest/kotlin/mu/KotlinLoggingTest.kt
@@ -1,18 +1,20 @@
 package mu
 
+import mu.KotlinLogging.logger
 import org.junit.Assert.assertEquals
 import org.junit.Test
 import org.slf4j.LoggerFactory
 
-private val logger = KotlinLogging.logger { }
-private val loggerFromSlf4j = KotlinLogging.logger(LoggerFactory.getLogger("mu.slf4jLogger"))
+private val logger = logger { }
+private val loggerFromSlf4j = logger(LoggerFactory.getLogger("mu.slf4jLogger"))
 private val loggerFromSlf4jExtension = LoggerFactory.getLogger("mu.slf4jLoggerExtension").toKLogger()
 
-class ForKotlinLoggingTest {
-    val loggerInClass = KotlinLogging.logger { }
+private class ForKotlinLoggingTest {
+    val loggerInClass = logger { }
+    val inlineLoggerInClass = logger()
 
     companion object {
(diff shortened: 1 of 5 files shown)


oshai:
note there are conflicts with previous PR

src/jvmTest/kotlin/mu/KotlinLoggingTest.kt
```diff
@@ -1,18 +1,20 @@
 package mu
 
+import mu.KotlinLogging.logger
```
oshai, line 3: I think it makes things more confusing

src/jvmTest/kotlin/mu/KotlinLoggingTest.kt
```diff
@@ -1,18 +1,20 @@
...
 import org.junit.Assert.assertEquals
 import org.junit.Test
 import org.slf4j.LoggerFactory
 
-private val logger = KotlinLogging.logger { }
-private val loggerFromSlf4j = KotlinLogging.logger(LoggerFactory.getLogger("mu.slf4jLogger"))
+private val logger = logger { }
+private val loggerFromSlf4j = logger(LoggerFactory.getLogger("mu.slf4jLogger"))
 private val loggerFromSlf4jExtension = LoggerFactory.getLogger("mu.slf4jLoggerExtension").toKLogger()
```
oshai, line 11: can we have a test also when logger() is defined outside of the class?

severn-everett: I don't think it's possible to use an reified inline extension function at the top level - the compiler only recognizes the other three public `KotlinLogging.logger()` functions as valid. This declaration works:
```
private val inlineLogger = KotlinLogging.logger()
```
However, this is declaring a logger for `mu.KotlinLogging`, not for the class containing the top-level variables.

src/jvmTest/kotlin/mu/KotlinLoggingTest.kt
severn-everett: The function has to be imported directly in order for the reified inline function to be executed properly. If it were called using the object name (i.e. `val inlineLoggerInClass = KotlinLogging.logger()`) then the logger will be instantiated for `mu.KotlinLogging`, not for `mu.ForKotlinLoggingTest`.

oshai:
I think that given the limitations of the api shown in the PR, it's better not to merge it.

It's easy to use the new api in a way that will not work as expected.

WDYT?

severn-everett:
I agree, also because it's not really going to be worthwhile in JS because `qualifiedName` isn't allowed in KotlinJS.

PR #207, opened 2021-12-10
Fix critical vulnerability in log4j

stigkj:
There is a critical vulnerability in Log4j, more information here:
https://www.lunasec.io/docs/blog/log4j-zero-day/

### DIFF
--- a/versions.gradle.kts
+++ b/versions.gradle.kts
@@ -1,4 +1,4 @@
 extra["slf4j_version"] = "1.7.32"
-extra["log4j_version"] = "2.14.1"
+extra["log4j_version"] = "2.15.0"
 extra["mockito_version"] = "1.10.19"
 extra["junit_version"] = "5.8.2"


hudson155:
LGTM – thanks for doing this

oshai:
Thanks for the PR!

edeak:
@hudson155  do you guys have an ETA for rolling this out to a release?

oshai:
I am releasing 2.1.16 with the change now. Note that since this is only a test dep it's not transitive so need really needed by users of the lib.

nippip:
> I am releasing 2.1.16 with the change now. Note that since this is only a test dep it's not transitive so need really needed by users of the lib.

@oshai Would you mind elaborating on your statement above and point to some code snips to better understand how this is or is not affecting users of the lib?

oshai:
> > I am releasing 2.1.16 with the change now. Note that since this is only a test dep it's not transitive so need really needed by users of the lib.
> 
> @oshai Would you mind elaborating on your statement above and point to some code snips to better understand how this is or is not affecting users of the lib?

`kotlin-logging` is a facade (similar to slf4j) means it only wraps the implementation you choose as a logging framework. 
 Popular implementations are logback and log4j.
On the tests of kotlin-logging we needed an actual implementation to use. We could write our own mocks but it makes more sense to take an actual impl for that to reflect how users use the lib. So kotlin-logging depends on log4j for testing only. 
Example of such tests can be seen [here](https://github.com/MicroUtils/kotlin-logging/blob/master/src/jvmTest/kotlin/mu/LoggingWithLocationTest.kt).
You can see the dependency [here](https://github.com/MicroUtils/kotlin-logging/blob/981daeae64780405092dab2420d4df342e01342b/build.gradle.kts#L71):
image

It means that this dependency does not pass transitively to users of kotlin-logging, and therfore users don't need to upgrade kotlin-logging itself. You can also see that in the [pom](https://repo1.maven.org/maven2/io/github/microutils/kotlin-logging-jvm/2.1.15/kotlin-logging-jvm-2.1.15.pom) that is generated for kotlin-logging.  
For people to feel more safe and for our own safety (when running tests) we upgraded kotlin-logging dependency on log4j.  
So it's not really needed, but I will not stop you from upgrading.

Hope that clears things a bit more.

PR #217, opened 2021-12-19
refactoring

BorzdeG:
* Added [Gradle version catalog](https://docs.gradle.org/current/userguide/platforms.html)
* Added shared IDEA configuration (minimal requirements)
* Upgrading Gradle plugins versions
* Upgrading Gradle version
* refactoring Gradle Actions
* Raised minimal Kotlin version from 1.4 to 1.7


oshai:
Thanks for the PR!
Any reason for changing kotlin min version from 1.4 to 1.5?
In addition, can you give some TL;DR about the catalog?

BorzdeG:
> Any reason for changing kotlin min version from 1.4 to 1.5?
https://kotlinlang.org/docs/whatsnew15.html#stable-jvm-ir-backend

> Starting from Kotlin 1.4.0, early versions of the IR-based backend were available for preview, and **_it has now become the default for language version 1.5_**. The old backend is still used by default for earlier language versions.

I want to turn off Legacy and leave only IR ...

BorzdeG:
> * Each catalog is visible to all projects of a build. It is a central place to declare a version of a dependency and to make sure that a change to that version applies to every subproject.
> * Catalogs can declare dependency bundles, which are "groups of dependencies" that are commonly used together.
> * Catalogs can separate the group and name of a dependency from its actual version and use version references instead, making it possible to share a version declaration between multiple dependencies.

+ rich versions support : https://docs.gradle.org/current/userguide/platforms.html#sub::toml-dependencies-format

oshai:
This PR is marked as WIP. let me know if it's ready for review.

yeikel:
@BorzdeG Are you planning to resume this PR in the near future?

oshai:
@BorzdeG the issue was closed and reoped. Is it still WIP?

PR #227, opened 2022-04-09
Add support for os_log on Apple platforms

conradev:
Apple platforms (macOS, iOS, watchOS, tvOS) have a [unified logging system](https://developer.apple.com/documentation/os/logging?language=objc) called `os_log`. I implemented support for `os_log` and made it the default on Apple platforms. This is similar to how `slf4j-android` implements `slf4j-api` to write to the Android system log.

I kept the console output appender still available on Apple platforms because it might still be desirable to use it in, for example, a command line tool.

`os_log` has two key concepts – a subsystem and a category. A subsystem is intended to be constant for an entire project (i.e. the reverse DNS notation for the project) and the category is supposed to change based on the component. Neither are usually part of the log message string itself, but rather used for filtering the log messages. If you have a subsystem and a category, you can create a log object with [`os_log_create(2)`](https://developer.apple.com/documentation/os/1643744-os_log_create), but if you don't have either you can just use [`OS_LOG_DEFAULT`](https://developer.apple.com/documentation/os/os_log_default)

My implementation makes `OS_LOG_DEFAULT` the default for Apple platforms, but you can specify a subsystem with:

```
KotlinLoggingConfiguration.appender = OSLogSubsystemAppender("com.example.app")
```

If you are using the default log, the `loggerName` is logged as a part of the message, but if you are using a specific subsystem, the `loggerName` will instead be used as the category. Similar to the subsystem and the category `os_log` also makes the level a parameter of the message itself (and something you can filter on), instead of as a part of the message text.

I hide the log level and `loggerName` if you are using a `OSLogSubsystemAppender`, but this part of the diff is a bit rough, and I'd love your thoughts on which direction it should go. For the time being, it works for my needs :)

`os_log` also has the ability to disable log levels for performance, and my implementation supports that as well.

Thank you for writing this library, it is working great for me! Feel free to suggest changes, or to just make them yourself – the branch is yours 😊

One small note is that Kotlin multiplatform handles the artifact multiplexing automatically – I am able to just specify the single `io.github.microutils:kotlin-logging` dependency in my `commonMain` and It Just Works™

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -51,9 +51,24 @@ kotlin {
         nodejs()
     }
 
-    linuxX64("linuxX64")
-    macosX64("macosX64")
-    mingwX64("mingwX64")
+    val linuxTargets = listOf(
+        linuxArm64(),
+        linuxX64(),
+        mingwX64()
+    )
+    val darwinTargets = listOf(
+        macosArm64(),
+        macosX64(),
+        iosArm64(),
+        iosSimulatorArm64(),
+        iosX64(),
+        watchosArm64(),
+        watchosSimulatorArm64(),
+        watchosX64(),
+        tvosArm64(),
+        tvosSimulatorArm64(),
+        tvosX64()
+    )
 
     sourceSets {
         val commonMain by getting {}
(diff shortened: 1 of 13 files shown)


oshai:
Thanks for the PR! I will look into it as soon as possible.

oshai:
LGTM!
Do you want to also create a documentation for that (in the wiki?) once merged?

conradev:
Yes, I can do that! I will also make sure that the README is updated in the commit before merge

oshai:
Thanks for the PR!

PR #240, opened 2022-09-11
Modify the contents of the README

juno-choi:
README url modify


oshai:
The link is like that on purpose. I think it is confusing to create a link that is not matching the text.

yeikel:
This link should probably be the title of the page

PR #238, opened 2022-08-28
Bump gradle wrapper to 7.5.1

yeikel:

PR #285, opened 2023-02-21
ci: add Gradle wrapper validation

yeikel:
This will help to validate changes like #238 

See https://github.com/gradle/wrapper-validation-action and https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification


oshai:
Thanks!

PR #291, opened 2023-03-25
build(deps): add dependabot configuration

yeikel:
See https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

This should reduce the manual work needed to create pull requests such as https://github.com/oshai/kotlin-logging/pull/287

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"


.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
yeikel, line 6: I am leaving it up to you to set the `interval` that works best for you. 

Options include: daily, weekly, monthly

I think that `weekly` is a good default but we can adjust as needed

PR #313, opened 2023-05-18
Allow turning off logging with `Level.OFF`

### LINKED ISSUE #310: Support turning off logging in platforms other than the JVM
I'm currently using version `4.0.0-beta-27`.

The current implementation seems to make it impossible to turn off logging in JS (and possibly all other platforms except JVM) since `KotlinLoggingConfiguration.LOG_LEVEL` is of type `Level` and there seems to be no way of configuring the logging level via integer.

I suggest the introduction of a new level: `Level.OFF`.

For reference, `sl4j-simple` provides an `OFF` level, defined as:

```java
// The OFF level can only be used in configuration files to disable logging.
// It has no printing method associated with it in o.s.Logger interface.
protected static final int LOG_LEVEL_OFF = LOG_LEVEL_ERROR + 10;
```

My current use-case for disabling logging is in unit testing. I wish to temporarily disable logging during certain tests so that I can test error scenarios without "polluting" the console. I'm working on a Kotlin multiplatform library and using `sl4j-simple` together with the default JS logger in unit testing.

YarnSphere:
Fixes #310.

I added the minimum amount of code to make this work, together with a test for JS and native.

It might make sense to add an `isOff` property to `KLogger` (which could probably even be implemented in common code as `!isErrorEnabled`), let me know what you think.

### DIFF
--- a/src/androidMain/kotlin/io/github/oshai/KLogger.kt
+++ b/src/androidMain/kotlin/io/github/oshai/KLogger.kt
@@ -118,4 +118,11 @@ public actual interface KLogger {
    * @return True if this Logger is enabled for the ERROR level, false otherwise.
    */
   public actual val isErrorEnabled: Boolean
+
+  /**
+   * Is the logger instance OFF?
+   *
+   * @return True if this Logger is set to the OFF level, false otherwise.
+   */
+  public actual val isLoggingOff: Boolean
 }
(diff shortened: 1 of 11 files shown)


oshai:
the PR lgtm thanks!
Adding a method sounds good. Maybe call is `isLoggingOff`?

YarnSphere:
Got it, I'll give it a go when I have a few more minutes to spare. Feel free to push to this branch if you manage to get to it before I do.

oshai:
Thanks!
Why didn't you implement it as `!isErrorEnabled` in the common interface? Isn't it enough?
Sounds like a good default.

oshai:
Thanks! Since I am doing other changes in #320 I am going to merge this one before to avoid conflicts.

Might change some things that were discussed in previous comments later.

PR #331, opened 2023-07-03
Append entries of payload to LoggingEventBuilder

imherrera:
payload entries should be added to slf4j builder so implementations like logback can make use of this data


oshai:
Thanks!
Would you like to add a test for it here or on another cl?

PR #347, opened 2023-08-06
add payload to formatter

oshai:


robertfmurdock:
x-posted from issue:

Makes sense to me. Would it make sense for the message to be passed through as the simple string then instead of the function arg? Or both of them be passed as function args? It seems slightly inconsistent for one to be a function producer and the other a flat map, but if there's an underlying mechanical reason that'd be fine

oshai:
Fixed in https://github.com/oshai/kotlin-logging/pull/350

PR #348, opened 2023-08-08
change darwin impl to delegate to os_log_t

oshai:
based on previous impl and https://blog.shipbook.io/ios-logs

looks like a better approach will be to just delegate logs to the os logger.


oshai:
@conradev following a refactoring around multiplatform I thought about simplifying darwin support.
As the author of the original PR #227 I would really appreciate to get your opinion.

This PR is changing darwin to behave more similar to other platforms that delegate logging to the native system. Major changes are:
- console logging is not supported anymore by itself (as long as the os logging don't support that).
- the resolution of subsystem and category was changed as follows:
`KotlinLoggingConfiguration` contains two optional fields:
```
public object KotlinLoggingConfiguration {
  public var subsystem: AtomicReference<String?> = AtomicReference(null)
  public var category: AtomicReference<String?> = AtomicReference(null)
}
```
- If both set they will be used as constant logger.
- if one is set it will be used, and the configured logger name will be used as the other.
- if both are null and logger name is empty `OS_LOG_DEFAULT` will be used.
- lastly if name contains "." it will be split and the category will be the part after the dot.
- and finally if name does not contain dot will result in blank category and logger name as  subsystem.

The code can be seen in `KLoggerFactory`.

I would really appreciate any thought about the revised impl. If console logging is something often used we can consider of a way to have both the new impl and console logging (with a config to pick which is used). But most other platforms that have clear native logging don't have that.

src/darwinMain/kotlin/io/github/oshai/kotlinlogging/DarwinKLogger.kt
```diff
@@ -0,0 +1,43 @@
...
+import platform.darwin.os_log_type_enabled
+import platform.darwin.os_log_type_t
+
+public class DarwinKLogger(override val name: String, override val underlyingLogger: os_log_t) :
+  KLogger, DelegatingKLogger<os_log_t> {
+
+  override fun at(level: Level, marker: Marker?, block: KLoggingEventBuilder.() -> Unit) {
+    if (isLoggingEnabledFor(level, marker)) {
+      KLoggingEventBuilder().apply(block).run {
+        _os_log_internal(__dso_handle.ptr, OS_LOG_DEFAULT, level.toDarwinLevel(), message)
```
conradev:
```suggestion
        _os_log_internal(__dso_handle.ptr, underlyingLogger, level.toDarwinLevel(), message)
```
I think this might need fixin

conradev (approved):
This overall approach looks great to me! I am glad that it is more standard for the library. I only started with console appender because that is what the existing macOS support used and I didn't want to rock the boat

I noted a small bug, but besides that it looks great!

conradev:
Next up, you should consider defaulting to journalctl on Linux if it is present if you don't already

oshai:
> Next up, you should consider defaulting to journalctl on Linux if it is present if you don't already

Thanks for the quick review! Regarding linux, an initial contribution is welcome (mostly how to access the linux api) and I can take if from there if you would like.

PR #379, opened 2023-12-05
Log marker and stacktrace on Darwin

### LINKED ISSUE #375: Darwin doesn't log stacktraces
`Throwable`s passed into logging functions on Darwin platforms will not print a stacktrace. There is simply no code that uses the `KLoggingEventBuilder.cause` on Darwin. This could potentially be resolved with Kotlin's `Throwable.stackTraceToString()`, which is available on Darwin (and all platforms).

Gaming32:
Resolves #375


oshai:
There are some failures in CI, also if you can add a test that would be great.

Gaming32:
I fixed Spotless, but I don't really know how I can add a test for this.

PR #384, opened 2023-12-22
Update Detekt to 1.23.4

AzimMuradov:
Should fix issues in #374 automatic PR

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,5 +9,8 @@ max_line_length = 150
 indent_style = space
 charset = utf-8
 
+[*.kt]
+indent_size = 2
+
 [*.yml]
 indent_size = 2
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 import io.gitlab.arturbosch.detekt.Detekt
+import io.gitlab.arturbosch.detekt.DetektCreateBaselineTask
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat.FULL
 import org.gradle.jvm.tasks.Jar
 import org.jetbrains.kotlin.gradle.ExperimentalKotlinGradlePluginApi
@@ -10,7 +11,7 @@ plugins {
     `maven-publish`
     id("io.github.gradle-nexus.publish-plugin") version "1.3.0"
     signing
-    id("io.gitlab.arturbosch.detekt") version "1.18.0"
+    id("io.gitlab.arturbosch.detekt") version "1.23.4"
     id("com.android.library") version "7.4.2"
     id("com.diffplug.spotless") version "6.23.3"
 }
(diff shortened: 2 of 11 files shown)


AzimMuradov:
Changes:

- Update Detekt to 1.23.4
  - Update detekt config

  Updated according to the detekt README (https://github.com/detekt/detekt?tab=readme-ov-file#with-gradle).

  - Add `formatting` detekt plugin
  
  In the newest detekt versions, the `formatting` rule set should be added separately via `detektPlugins` mechanism.
  
  From https://github.com/detekt/detekt?tab=readme-ov-file#adding-more-rule-sets:
  
  > detekt itself provides a wrapper over [ktlint](https://github.com/pinterest/ktlint) as the formatting rule set which can be easily added to the Gradle configuration:
  >
  > ```kts
  > dependencies {
  >     detektPlugins("io.gitlab.arturbosch.detekt:detekt-formatting:[version]")
  > }
  > ```

  - Clean up the code to pass the updated checks

- Reorganize build.gradle.kts

I hope this change would make the build script more readable and generally nice to config.

- Set up indentation level to 2 spaces

This style guide is already present in the project, so I decided to explicitly describe it in the .editorconfig and detekt.yml.

AzimMuradov:
BTW, why does the build script use `spotless` and `detekt` at the same time? I thought the `detekt` would be more than enough.

AzimMuradov:
Also, I could filter out detekt tasks on `...Test` source sets if it's needed.

It's as simple as changing

```kts
tasks {

    // ...

    afterEvaluate {
        check {
            dependsOn(withType<Detekt>())
        }
    }
}
```

to

```kts
tasks {

    // ...

    afterEvaluate {
        check {
            dependsOn(withType<Detekt>().filterNot { it.name.contains("test", ignoreCase = true) })
        }
    }
}
```

oshai:
> BTW, why does the build script use `spotless` and `detekt` at the same time? I think the `detekt` would be more than enough.

If I remember correctly we started with detekt and now use spotless (don't remember why I changed it...). So I am not sure we need detekt at all.

AzimMuradov:
> > BTW, why does the build script use `spotless` and `detekt` at the same time? I think the `detekt` would be more than enough.
> 
> If I remember correctly we started with detekt and now use spotless (don't remember why I changed it...). So I am not sure we need detekt at all.

As far as I know, detekt is a full blown static analysis tool while spotless is more like a formatter/code style checker.

If you want, I can remove detekt from the config. If so, should I just rename this pr, or should I close it, and open the new one?

AzimMuradov:
The reason I created this PR is because old `detekt` blocks the addition of the `wasmJs` target. So, if you currently don't want to remove `detekt`, we can think about it later and I'd just update it version for now. If you don't currently want to remove `detekt` we can think about that later and for now I'll just update it's version.

oshai:
Let's stay only with spotless and remove detekt.

Issue #306, opened 2023-04-19
[feature-request] support wasm target


VladislavSumin:
It is possibly to support wasm target? Coroutines and Compose libraries already have experimental wasm target support

oshai:
If you have any reference how to add such support to a lib I would be happy to add it. Of course contribution is also welcome.

LaylaMeower:
> If you have any reference how to add such support to a lib

WASM is a compile target, that similarly to JS, has `browser` and `nodejs` - but it also has `d8` (no idea what that is tbh).

WASM Support will be very appreciated!

oshai:
If I understand correctly, our current "native" implementation (writing logs using println) can also work in wasm. So the main thing to add is the gradle configuration. No need for many code changes.

LaylaMeower:
Do note that when WASM is out of alpha, tit'll likely have interop with other WASM languages, allowing for usage of more complex log libraries than the basic stdout.

Plus, you can use JS `console` in WASM - it has JS interop.

oshai:
Which impl do you think is the most appropriate?

LaylaMeower:
> Which impl do you think is the most appropriate?

I think `console`, as it supports multiple log levels. *However*, `println` delegates to the standard WASM output. To allow a more more smooth interop with other languages when the time comes, I would suggest the option to have both, by making one the default and the other an add-on artifact (or making both an add-on artifact each)

The same, btw, applies to `println` - if you choose that, interop with JS will be less smooth.

(`console` supports, iirc, `debug`, `normal`/`info`, `warning`, and `error`, while `println` is only `normal`/`info` and `error`)

gsteckman:
Kotlin for WebAssembly is now [alpha](https://blog.jetbrains.com/kotlin/2023/12/kotlin-for-webassembly-goes-alpha/).

severn-everett:
What's left for this issue?

oshai:
I am marking it as completed.

PR #388, opened 2024-01-14
Add Kotlin/Wasm (JS browser) support

AzimMuradov:
Partially solves #306.

No Node.js support for now, as it requires a canary Node.js version ([21.0.0-v8-canary202309143a48826a08 or newer](https://youtrack.jetbrains.com/issue/KT-63014/Running-tests-with-wasmJs-in-1.9.20-requires-Chrome-Canary#focus=Comments-27-8321383.0-0)).

No WASI support also. Considering the raw state of Kotlin WASI, this currently seems too hard to implement.

Overall, this is a fairly simple implementation based on the `jsMain` and `jsTest` modules. I hope this will be ok as a first solution, just to provide initial support for WASM.

### DIFF
--- a/src/wasmJsMain/kotlin/io/github/oshai/kotlinlogging/ConsoleOutputAppender.kt
+++ b/src/wasmJsMain/kotlin/io/github/oshai/kotlinlogging/ConsoleOutputAppender.kt
@@ -1,0 +1,30 @@
+package io.github.oshai.kotlinlogging
+
+public class ConsoleOutputAppender : FormattingAppender() {
+  override fun logFormattedMessage(loggingEvent: KLoggingEvent, formattedMessage: Any?) {
+    when (loggingEvent.level) {
+      Level.TRACE -> consoleLog(formattedMessage.toString())
+      Level.DEBUG -> consoleLog(formattedMessage.toString())
+      Level.INFO -> consoleInfo(formattedMessage.toString())
+      Level.WARN -> consoleWarn(formattedMessage.toString())
+      Level.ERROR -> consoleError(formattedMessage.toString())
+      Level.OFF -> Unit
+    }
+  }
+}
+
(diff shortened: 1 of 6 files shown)


src/wasmJsMain/kotlin/io/github/oshai/kotlinlogging/ConsoleOutputAppender.kt
```diff
@@ -0,0 +1,30 @@
...
+      Level.INFO -> consoleInfo(formattedMessage.toString())
+      Level.WARN -> consoleWarn(formattedMessage.toString())
+      Level.ERROR -> consoleError(formattedMessage.toString())
+      Level.OFF -> Unit
+    }
+  }
+}
+
+private fun consoleLog(
+  @Suppress("UNUSED_PARAMETER") msg: String,
```
oshai, line 17: can you explain why it's not used? Anyway I suggest to change it here to message and keep tests with msg so it's clear if it's really logging the input.

oshai (commented):
Thanks!

src/wasmJsMain/kotlin/io/github/oshai/kotlinlogging/ConsoleOutputAppender.kt
AzimMuradov: > can you explain why it's not used?

It's used, but in a weird way (in the `js("console.log(msg)")` expression). Unfortunately, linters and IDEs sometimes don't understand this syntax, so I've suppressed potential warnings.

AzimMuradov: I renamed `msg` to `message` and removed annotations :)

oshai:
Thank you for the contribution!

PR #389, opened 2024-01-14
Update README.md

sksamuel:
Added https://github.com/grindrllc


oshai:
Hi, where is kotlin-logging used in https://github.com/grindrllc?

sksamuel:
internal repos. If you only want open source links, then we can close this sorry !

oshai:
It doesn't have to be open source only, but it have to be stated some where that kotlin-logging is used.

sksamuel:
Ok makes sense. Sorry for the confusion.

PR #394, opened 2024-01-15
extract DarwinFormatter and add test

oshai:


deejay1 (approved):

oshai:
For some reason the test here do not run (it should fail).

Issue #400, opened 2024-02-24
KLoggingEventBuilder accept nullable values in payload


Zack-Freedman-Thoughtworks:
Hello! Is it possible to modify KLoggingEventBuilder to accept nullable values? This would allow us to avoid null-checks in the code we are adding logs to. 

Example:
```Kotlin
logger.atError {
    this.message = "Failed to make a good example: ${throwable.message}."
    this.cause = throwable
    this.payload = mapOf(
        "itemOne" to headers.getNullableThing(),
    )
}
```

oshai:
yes. would you like to create a PR?

Zack-Freedman-Thoughtworks:
@oshai Created a PR [here](https://github.com/oshai/kotlin-logging/pull/402). Would you please help me validate that it doesn't lead to a null-pointer? I'm not too knowledgable in how this library integrates with the other libraries as a whole. 

Sadly as these are public classes, I'm not sure if these are backwards compatible changes. I also happen to know data classes don't fair well with backwards compatible changes either. See [here](https://kotlinlang.org/docs/jvm-api-guidelines-backward-compatibility.html)

oshai:
I think the change is safe (both NPE and backward compatibility).

PR #402, opened 2024-02-27
Fluent Logging: Allow nullable values

Zack-Freedman-Thoughtworks:
References: https://github.com/oshai/kotlin-logging/issues/400#issuecomment-1964617668

Example:
```kotlin
Example:

logger.atError {
    this.message = "Failed to make a good example: ${throwable.message}."
    this.cause = throwable
    this.payload = mapOf(
        "itemOne" to headers.getNullableThing(),
    )
}
```

Allows the value in the payload to be nullable. It seems that the code referencing the payloads are are written in Java so it is harder to know if this will lead to a null pointer exception or other troubles with possible loggers but the tests pass.


oshai:
From code checks I did I think the Java frameworks allow values to be null.

oshai:
Thanks!

PR #408, opened 2024-03-22
allow only markers being supplied

ewoelfel:
I think the api could be improved a bit by not requiring the throwable, which could result in the following:
```
logger.info(marker=myMarker) { "some Message" }
```
Especially you do not need to supply a throwable when logging info in this case.
Due to the deprecation of the previous methods the test still uses deprecated methods. 
If the old methods are removed, the problem wont occur anymore.


oshai:
LGTM, thanks!

PR #414, opened 2024-04-01
Allow supplying org.slf4j.Marker

### LINKED ISSUE #360: Problem using LogstashMarker
Hi @oshai!
First of all, thank you for your work in the library!

I'm having an issue migrating to v5 and I can't seem to find a good approach to solve this problem.

We are using `LogstashMarker`s which is basically an implementation of `org.slf4j.Marker` interface provided by the `logstash-logback-encoder` library. 
Migrating to V5 from V3 is proving to be more difficult because the `KLogger` methods only accept a `io.github.oshai.kotlinlogging.Marker` interface. 

In this case, how would you suggest to proceed? I tried looking into the documentation and other issues in the project but could not find any approach to this.

Thank you for time!

mlewe:
This fixes #360.

~~`./gradlew spotlessCheck` fails on this PR because of the ancient version of ktfmt used in the project (0.24).~~

### DIFF
--- a/src/javaMain/kotlin/io/github/oshai/kotlinlogging/slf4j/Slf4jExtensions.kt
+++ b/src/javaMain/kotlin/io/github/oshai/kotlinlogging/slf4j/Slf4jExtensions.kt
@@ -5,10 +5,17 @@ import io.github.oshai.kotlinlogging.KotlinLogging
 import io.github.oshai.kotlinlogging.Level
 import io.github.oshai.kotlinlogging.Marker
 import io.github.oshai.kotlinlogging.slf4j.internal.Slf4jLoggerFactory
+import io.github.oshai.kotlinlogging.slf4j.internal.Slf4jMarker
 import org.slf4j.Logger
 import org.slf4j.MarkerFactory
 
-public fun Marker.toSlf4j(): org.slf4j.Marker = MarkerFactory.getMarker(this.getName())
+public fun Marker.toSlf4j(): org.slf4j.Marker =
+  when (this) {
+    is Slf4jMarker -> marker
+    else -> MarkerFactory.getMarker(getName())
+  }
+
+public fun org.slf4j.Marker.toKotlinLogging(): Marker = Slf4jMarker(this)
 
 public fun Level.toSlf4j(): org.slf4j.event.Level =
   when (this) {
--- a/src/javaMain/kotlin/io/github/oshai/kotlinlogging/slf4j/internal/Slf4jMarker.kt
+++ b/src/javaMain/kotlin/io/github/oshai/kotlinlogging/slf4j/internal/Slf4jMarker.kt
@@ -1,0 +1,5 @@
+package io.github.oshai.kotlinlogging.slf4j.internal
+
+import io.github.oshai.kotlinlogging.Marker
+
+internal class Slf4jMarker(val marker: org.slf4j.Marker) : Marker, org.slf4j.Marker by marker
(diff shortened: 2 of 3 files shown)


src/javaMain/kotlin/io/github/oshai/kotlinlogging/slf4j/Slf4jExtensions.kt
```diff
@@ -25,3 +30,150 @@ public fun KotlinLogging.logger(underlyingLogger: Logger): KLogger =
   Slf4jLoggerFactory.wrapJLogger(underlyingLogger)
 
 public fun Logger.toKLogger(): KLogger = KotlinLogging.logger(this)
+
+/** Lazy add a log message if isTraceEnabled is true */
+public fun KLogger.trace(
+  throwable: Throwable? = null,
+  marker: org.slf4j.Marker?,
```
oshai, line 37: Maybe I missed something, but why those methods needs to be added?
Anyway Marker.toSlf4j() will treat it correctly when the actual call happens, isn't it?

mlewe: I've added those extension methods as a convenience, so the user doesn't have to convert their `org.slf4j.Marker` to `io.github.oshai.kotlinlogging.Marker` manually.

If you think this is not desired, I can remove them from this PR.

What essentially happens is `marker.toKotlinLogging().toSlf4j()` which is equivalent to `marker` for the call to the underlying Logger.
(And if the underlying Logger is not Slf4j, it still behaves as a normal marker using the implementation of `getName()` from `org.slf4j.Marker`.)

oshai: got it. I think I prefer to keep the api slim, not adding those extension methods. So it's better to remove from the PR.

mlewe: I've updated the PR accordingly.

src/javaMain/kotlin/io/github/oshai/kotlinlogging/slf4j/Slf4jExtensions.kt
```diff
@@ -1,14 +1,18 @@
 package io.github.oshai.kotlinlogging.slf4j
 
-import io.github.oshai.kotlinlogging.KLogger
-import io.github.oshai.kotlinlogging.KotlinLogging
-import io.github.oshai.kotlinlogging.Level
-import io.github.oshai.kotlinlogging.Marker
+import io.github.oshai.kotlinlogging.*
```
oshai, line 3: please don't use star import

src/javaMain/kotlin/io/github/oshai/kotlinlogging/slf4j/internal/Slf4jMarker.kt
```diff
@@ -0,0 +1,6 @@
+package io.github.oshai.kotlinlogging.slf4j.internal
+
+import io.github.oshai.kotlinlogging.Marker
+
+@JvmInline
```
oshai, line 5: I am not sure there is a real performance gain here from inline. it might even be the opposite because of boxing / unboxing. So unless there is an obvious reason I suggest to have it a regular class.

oshai:
Thanks!

mlewe:
Thank you and thanks for the feedback :)

PR #428, opened 2024-05-23
Upgraded Kotlin to 2.0.0

severn-everett:


build.gradle.kts
```diff
@@ -42,19 +46,13 @@ kotlin {
     @OptIn(ExperimentalKotlinGradlePluginApi::class)
     compilerOptions {
         // kotlin compiler compatibility options
-        apiVersion.set(KotlinVersion.KOTLIN_1_9)
-        languageVersion.set(KotlinVersion.KOTLIN_1_9)
+        apiVersion.set(KotlinVersion.KOTLIN_2_0)
+        languageVersion.set(KotlinVersion.KOTLIN_2_0)
 
         freeCompilerArgs.add("-Xexpect-actual-classes")
```
oshai, line 52: is this still required?

severn-everett:
Btw if you've got any idea about how to investigate why the `kotlin-stdlib-js:1.9.23` library is cached in Gradle plugin repository and not the `kotlin-stdlib-js:2.0.0` library, I'd appreciate the guidance, because requiring the Maven repository declaration to come first seems fragile as a solution.

oshai:
> Btw if you've got any idea about how to investigate why the `kotlin-stdlib-js:1.9.23` library is cached in Gradle plugin repository and not the `kotlin-stdlib-js:2.0.0` library, I'd appreciate the guidance, because requiring the Maven repository declaration to come first seems fragile as a solution.

Maybe open an issue for the kotlin team to investigate with a reproduction?

severn-everett:
Based on my [consultation](https://youtrack.jetbrains.com/issue/KT-68471) with the Kotlin devs, I suspected that this might have been an issue with IntelliJ, so I re-ran the build after upgrading to IntelliJ 2024.1.2, and it worked.

However, reverting the repository declaration change in the code caused the build to fail in my forked version in a similar manner: https://github.com/severn-everett/kotlin-logging/actions/runs/9224844785

Yet it didn't fail in _this_ project's Github actions: https://github.com/oshai/kotlin-logging/actions/runs/9224845625

Any thoughts?

oshai:
In your fork I saw a different error related to wasm.

severn-everett:
It was the same issue of "Gradle can't find a Kotlin 2 multiplatform dependency in the Gradle Plugins repository", and it's shown up in the [Dependabot PR](https://github.com/oshai/kotlin-logging/pull/430) for updating to Kotlin 2 as well. However, I just re-ran the Github actions in my own fork successfully, so either it's a caching issue with Gradle or maybe some race condition. I'll ask the Kotlin devs, but this PR should be good to go now at least.

severn-everett:
According to the [Kotlin devs](https://youtrack.jetbrains.com/issue/KT-68471/KMP-Cannot-Find-kotlin-stdlib-js-when-mavenCentral-not-declared-first#focus=Comments-27-9851285.0-0), `gradlePluginPortal()` isn't required in any case, so I've removed it in my PR. Do we know for sure whether the declaration is necessary all the same?

oshai:
Thanks!
When merging the PR I would like to create a compatibility matrix in the readme as it seems we bumped few Kotlin versions recently.

oshai:
Thanks!

oshai:
My plan is to release this as version 7.

Issue #449, opened 2024-10-14
KLogger does not properly populate SLF4J/Logback ILoggingEvent "message" field


neeme-praks-sympower:
### Context
SLF4J/Logback `ILoggingEvent` ([source](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/spi/ILoggingEvent.java)) has two important fields / getter methods:
* `message` - contains the message template (e.g. `"Hello {}!"`)
* `formattedMessage` - contains the final formatted/merged log message (e.g. `"Hello world!"`)

This distinction is useful -- for example, we use this to aggregate log messages.

### Issue
`KLogger` relies on Kotlin string interpolation and as a result, the `message` field contains the same formatted message as `formattedMessage`, breaking the possibilities for aggregation.

### Proposal

Introduce a new field in `KLoggingEventBuilder` that could contain the log message template (e.g. `"Hello $variable!"`):
* in case this field is filled, it is passed as-is to SLF4J `LoggingEvent.message` field and the formatted log message is passed to `LoggingEvent.formattedMessage` (see also https://github.com/qos-ch/slf4j/issues/437).
* in case this field is not filled, the current behaviour is followed (using the formatted message for `LoggingEvent.message` and leaving `LoggingEvent.formattedMessage` empty).

This will make it easier for me to implement something on top of `KLogger` to fill that field properly (I'm thinking of implementing a Kotlin compiler plugin).

oshai:
Can you please give some more info?
You mean that the un-formatted message will be added by the compiler plugin? if so, sounds like an interesting idea. In the past I thought of using a compiler plugin to add line numbers, file name, etc'. But never implemented it.

neeme-praks-sympower:
> un-formatted message will be added by the compiler plugin?

Exactly.

> In the past I thought of using a compiler plugin to add line numbers, file name, etc'

Also possible, yes.

oshai:
ok, would you like to send a PR for that or let me do the proposal?

neeme-praks-sympower:
I'm mostly interested in the unformatted message. In case you have broader scope in mind then it is probably better if you come up with a proposal.

oshai:
I suggest that you'll do the initial PR for the unformatted message (maybe call it `rawMessage`) and I will add on top another PR. Make sense?

neeme-praks-sympower:
[...]

PR #452, opened 2024-10-24
Logback backend for kotlin-logging

neeme-praks-sympower:
Proposal for https://github.com/oshai/kotlin-logging/issues/449
I'm not sure if this change requires kotlin-logging to always have Logback on runtime classpath or not (how eager is JVM classloading nowadays?) -- but we can figure that detail out later.


oshai:
Right now it fails on formatting, see how to fix spotless here: https://github.com/oshai/kotlin-logging/blob/master/CONTRIBUTING.md

In addition, since I want it to be both extended (with `<class>, <method>, <line>, <file>`) maybe we should wrap it with `compilerOptions` or `internalCompilerData` data class. I think it will be more obvious that this should not be set.

neeme-praks-sympower:
Thanks for the feedback! I'll get back to this next week.

neeme-praks-sympower:
I've now fixed formatting issues and introduced `internalCompilerData` field.

src/commonMain/kotlin/io/github/oshai/kotlinlogging/KLoggingEventBuilder.kt
```diff
@@ -4,4 +4,13 @@ public class KLoggingEventBuilder {
   public var message: String? = null
   public var cause: Throwable? = null
   public var payload: Map<String, Any?>? = null
+  public var internalCompilerData: InternalCompilerData? = null
```
oshai, line 7: please document that this field is internal and should not be set

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/logback/internal/LogbackLoggerFactory.kt
```diff
@@ -0,0 +1,27 @@
...
+    val logbackServiceProvider = LogbackServiceProvider()
+    logbackServiceProvider.initialize()
+    return logbackServiceProvider
+  }
+
+  /** get a java logger by name. Logback relies on SLF4J logger factory */
+  internal fun jLogger(name: String): Logger =
+    logbackServiceProvider.loggerFactory.getLogger(name) as Logger
+
+  /** wrap java logger based on location awareness */
```
oshai, line 22: actually here there is no difference in location awerness so this can be removed

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/logback/internal/LogbackLogEvent.kt
```diff
@@ -0,0 +1,48 @@
...
+  fqcn: String,
+  logger: Logger,
+  level: Level,
+  private val kLoggingEvent: KLoggingEventBuilder,
+) :
+  LoggingEvent(
+    fqcn,
+    logger,
+    level.toLogbackLevel(),
+    kLoggingEvent.internalCompilerData?.messageTemplate ?: kLoggingEvent.message,
```
oshai, line 19: if you provide here message template, don't you need to also provide arguments?

oshai (commented):
lgtm, left some comments.

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/logback/internal/LogbackLogEvent.kt
neeme-praks-sympower: For my use case, I only need the template. And as the final log message has already been generated then there is no actual need for the arguments. So I would skip the arguments until someone comes and provides a real-world use-case where arguments are needed.

src/commonMain/kotlin/io/github/oshai/kotlinlogging/KLoggingEventBuilder.kt
neeme-praks-sympower: Added JavaDoc

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/logback/internal/LogbackLoggerFactory.kt
neeme-praks-sympower: Removed. And renamed methods and fields a bit.

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/logback/internal/LogbackLogEvent.kt
oshai: Isn’t slf4j such a case? In case we want to support it later?

neeme-praks-sympower: Well, if we eagerly merge message template with the arguments then SLF4J has no need for the arguments -- the message is final and no further processing is needed.

oshai:
Can you please fix spotless again before I submit it?

neeme-praks-sympower:
> Can you please fix spotless again before I submit it?

Fixed.

I would like to also include a test that would execute `JulLoggerWrapperTest` without Logback in the classpath -- to make sure that `KLoggerFactory` can access `JulLoggerFactory` without needing to load `LogbackLoggerFactory` (which depends on Logback).

I tried to play around with Gradle build file but my knowledge about Kotlin multiplatform (and Gradle) build system is limited -- maybe you can help out here?

oshai:
> > Can you please fix spotless again before I submit it?
> 
> Fixed.
> 
> I would like to also include a test that would execute `JulLoggerWrapperTest` without Logback in the classpath -- to make sure that `KLoggerFactory` can access `JulLoggerFactory` without needing to load `LogbackLoggerFactory` (which depends on Logback).
> 
> I tried to play around with Gradle build file but my knowledge about Kotlin multiplatform (and Gradle) build system is limited -- maybe you can help out here?

Thanks! I merged the PR and just now see this message. I think it will be difficult to do tests with / without logback in classpath. I also don't think gradle is very good at this. I think we can go without such test.

neeme-praks-sympower:
>  I think we can go without such test.

Just to be sure, I ran a manual test with a simple project on my machine -- seems to work just fine (using `-Dkotlin-logging-to-jul=true` and without Logback in the classpath).

Thanks!

Issue #457, opened 2024-11-25
Can't get extra fields in json output without relying on deprecated methods


sigmanil:
Hi, and thank you for maintaining an excellent library! 

We have unfortunately run into a little problem:

We noticed that the support for writing log statements on the form `log.error("Cake", StructuredArguments.keyValue("caketype", "chocolate"))` (with net.logstash.logback.argument.StructuredArguments) has been deprecated. We use that style to include extra fields for json logging.

The replace-with in the deprecation creates a wholly different result. After some searching, we noticed the "atError" etc function that take a payload, and hoped they could be used to replace the old behaviour.

```
log.atError {
   message = "Cake"
   payload = buildMap {
      put("caketype", "chocolate")
    }
}
```

But while that does include the caketype/chocolate pair in the output for ordinary logging, it disappears when we use Json logging.

For a fairly minimal example, see the below files and the comments in main. The behaviour of atError seems to be opposite of the behavior of the old method, in that the old one doesn't include the extra field in "ordinary" logging, while the new one does - and vice versa when it comes to json logging.

**Have I misunderstood what the payload argument of _atInfo_, _atWarn_ etc functions are supposed to be used for? If so, how do we reproduce the old behaviour for including extra fields in json output without leaning on the deprecated method?** (Or is this a bug?)

Thanks in advance for any hints or ideas, we're a bit stuck here. :-) 

## Main.kt
```
import io.github.oshai.kotlinlogging.KotlinLogging
import net.logstash.logback.argument.StructuredArguments

private val log = KotlinLogging.logger {}
object Main {
    @JvmStatic
    fun main(args: Array<String>) {
        log.error("Cake", StructuredArguments.keyValue("caketype", "chocolate"))
// Without json-config, prints: 15:56:38.134 [main] ERROR Main -- Cake
// With json-config, prints {"timestamp":"2024-11-25 15:54:30.553","level":"ERROR","message":"Cake","logger":"Main","class":"i.g.o.k.s.internal.Slf4jLogger","method":"error","line":"128","thread":"main","caketype":"chocolate"}


        log.atError {
            message = "Cake"
            payload = buildMap {
                put("caketype", "chocolate")
            }
        }
// Without json-config, prints 15:56:38.138 [main] ERROR Main -caketype="chocolate"- Cake
[...]

PR #458, opened 2024-11-27
Support arguments where possible for Logback and Slf4j

sigmanil:
Proposal for https://github.com/oshai/kotlin-logging/issues/457

Added arguments array and passed it to underlying loggers where possible, not all loggers seem to support it. (Array instead of the initially proposed Set, because that is the datatype used by the underlying loggers.)


oshai:
Thanks for the PR! Generally LGTM. Is it possible to add a unit test for your use case?

sigmanil:
Something like this? https://github.com/oshai/kotlin-logging/pull/458/commits/d57c87173b40efb32c0eb38fa093b785e16462a7

sigmanil:
Noticed that the previous commit failed on formatting (spotless checks). Fixed that now.

oshai:
Thanks! will be released in 7.0.1.

PR #479, opened 2025-02-27
Fix Logback integration -- use SLF4J API to get access to Logback logger context

neeme-praks-sympower:
### Fixed direct-to-Logback integration
(enabled with `kotlin-logging-to-logback` System property)

It was broken as it created a new instance of `LogbackServiceProvider` instead of getting Logback `LoggerContext` from  SLF4J `LoggerFactory`. As a result, the new `LogbackServiceProvider` was not properly initialised and log messages ended up on console (default initialisation logic fallback in Logback).

### Fixed deprecated SLF4J-like `KLogger` API throwing [NotImplementedError](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-not-implemented-error/index.html)

Some of the `KLogger` API methods throw [NotImplementedError](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-not-implemented-error/index.html) unless the `KLogger` implementation explicitly implements them. When used with [kotlin-logging compile-time plugin](https://github.com/nemecec/kotlin-logging-compile-time-plugin) calls to those methods are detected and rewritten to use non-deprecated API. However, when the plugin is not used then an error is thrown. This change adds implementation for those deprecated methods so logging works also without the plugin.

### Improved test-suite for direct-to-Logback integration

Logback tests and Log4J-with-SLF4J-over-JUL tests were interfering with each-other (this is also the original reason why a new instance of `LogbackServiceProvider` was used) so this change moves Logback tests to a separate compilation target (so those tests can have separate dependencies and are ran in isolation). Also, new tests were added to cover the newly implemented API methods.


neeme-praks-sympower:
@oshai, have you managed to have a look at this? When could we merge it in and release a new version?

oshai:
Yes, I will try to take a look soon.

src/jvmLogbackTest/kotlin/io/github/oshai/kotlinlogging/logback/internal/LogbackLoggerWrapperTest.kt
```diff
@@ -0,0 +1,421 @@
...
+              }
+            }
+            thenExpect { given ->
+              level = given.level
+              marker = given.marker
+              name = loggerName
+              message = "${given.messagePrefix} ${given.arg1} ${given.arg2}"
+            }
+          }
+          test("$levelValue SLF4J API with marker and varargs") {
```
oshai, line 346: do you think that those tests can be changed to regular test functions with annotations? I am not sure it will be easy to maintain those, and tools will have suboptimal experience with those. If it's for the multiple levels and given that it's unit tests, I am not sure we have to tests all levels for each scenario.

oshai (commented):
Thanks! 
Can you add some more details on the change in the PR description?

src/jvmLogbackTest/kotlin/io/github/oshai/kotlinlogging/logback/internal/LogbackLoggerWrapperTest.kt
neeme-praks-sympower: Good point, I simplified that complexity, have a look.

neeme-praks-sympower:
> Thanks! Can you add some more details on the change in the PR description?

I updated the description.

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerFactory.kt
```diff
@@ -12,10 +14,16 @@ internal actual object KLoggerFactory {
...
+    return if (LoggerFactory.getILoggerFactory().isLogback()) {
+      LogbackLoggerFactory.logbackLogger(name)
+    } else {
+      Slf4jLoggerFactory.wrapJLogger(Slf4jLoggerFactory.jLogger(name))
+    }
   }
 }
+
+private fun ILoggerFactory.isLogback() =
+  System.getProperty("kotlin-logging-to-logback") == "true" &&
```
oshai, line 28: I think it's better to have the property check outside, as it does not depend on class, and there are other property checks (jul) on the caller.

Also I think the extension method gives a fluent api, but otoh it doesn't really use the public properties of the subject, so it's less clear. If this will be inside the class, is it possible to check isInstance instead? - I mean in such case it should allow to import that class, isn't it?

oshai (commented):
Thanks for the previous fixes, added couple more comments.

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerFactory.kt
```diff
@@ -12,10 +14,16 @@ internal actual object KLoggerFactory {
   internal actual fun logger(name: String): KLogger {
     if (System.getProperty("kotlin-logging-to-jul") != null) {
       return JulLoggerFactory.wrapJLogger(JulLoggerFactory.jLogger(name))
-    } else if (System.getProperty("kotlin-logging-to-logback") != null) {
-      return LogbackLoggerFactory.wrapLogbackLogger(LogbackLoggerFactory.logbackLogger(name))
     }
-    // default to slf4j
-    return Slf4jLoggerFactory.wrapJLogger(Slf4jLoggerFactory.jLogger(name))
+    // Try Logback and if that fails then default to plain SLF4J
+    return if (LoggerFactory.getILoggerFactory().isLogback()) {
```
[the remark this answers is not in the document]
neeme-praks-sympower: Explanation: originally I implemented automatically enabling `LogbackLoggerFactory` when Logback is in use and then added the opt-in check for system property -- thus the extra complexity.
I now simplified the code again.

src/jvmMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerFactory.kt
neeme-praks-sympower: Explanation: originally I implemented automatically enabling `LogbackLoggerFactory` when Logback is in use and then added the opt-in check for system property -- thus the extra complexity.
I now simplified the code again.

neeme-praks-sympower:
Thanks! Deployed to production and works as expected! Grrreat!

Issue #480, opened 2025-03-03
Extend the way of how to create the logger


scrat98:
Note: I know that's an old discussion, but just to revise it again.

**Problem**

Currently, the "idiomatic" way to create logger is a top-level variable.

```kotlin
import io.github.oshai.kotlinlogging.KotlinLogging

// Place definition above class declaration, below imports, 
// to make field static and accessible only within the file
private val logger = KotlinLogging.logger {}
```

1) Lot's of JVM backend teams prefer to have private variables inside the class. What if we have couple of private static variables, should we also place them as a top level declaration? And what if we have instance private variables as well? This introduce inconsistency and hard-to-read code base, since some of the variables could be as a top-level, some of them as a companion object and some of them as a instance variables.

```kotlin
import io.github.oshai.kotlinlogging.KotlinLogging

private val logger = KotlinLogging.logger {}

class MyClass {
    private companion object {
        private val someStaticVar = 1
        private val anotherStaticVar = 2
    }

    private val someVar = 3
}
```

2) Let's add to the example above a combination with long doc on top of the class that's move out completely the logger from view.

```kotlin
import io.github.oshai.kotlinlogging.KotlinLogging

private val logger = KotlinLogging.logger {}

/**
 * very long Doc
 */
class MyClass {
    private companion object {
        private val someStaticVar = 1
        private val anotherStaticVar = 2
    }

    private val someVar = 3
}
```

3) Exposes the top-level class to the public API.
If I'm writing a Java library, adding a variable at the top level introduces an public empty top-level class. In recent version of Kotlin we can "embed" the top-level variable using `@file:JvmName`

```kotlin
@file:JvmName("MyPublicAPI") // may work in recent kotlin versions

package example

import io.github.oshai.kotlinlogging.KotlinLogging

private val logger = KotlinLogging.logger {}

class MyPublicAPI {

}
```

4) Top level variables introduce additional top level class (if we are not using `@file:JvmName` the same as class name). It just makes the final jar bigger.

5) Before Kotlin 2.0 each lambda functions compiled to the anonymous classes, that also makes final jar bigger, because for every logger we need a lambda. https://kotlinlang.org/docs/whatsnew20.html#generation-of-lambda-functions-using-invokedynamic

6) There is ongoing proposal about static namespace/static variables and current syntax does not allow to migrate (with IDEA/compiler hint) from this syntax:
```kotlin
class MyClass {
  private companion object {
    private val logger = KotlinLogging.logger(this)
  }
[...]

PR #481, opened 2025-03-03
Issue #480 - Extend the way of how to create the logger

scrat98:
Closes https://github.com/oshai/kotlin-logging/issues/480


oshai:
Thanks, I am not sure I understand this change. Why do we need to provide "this"? Can it be inferred somehow?

Also regardless - if we will add that method there is no plan to do a breaking change at the moment (removing the other methods).

PR #485, opened 2025-03-06
Issue #484 - Optimize the KLoggerNameResolver for jvm

### LINKED ISSUE #484: Optimize the KLoggerNameResolver for jvm
Part 1 of https://github.com/oshai/kotlin-logging/issues/480

Optimize the KLoggerNameResolver for jvm, since we don't need to search indexOf multiple times.

scrat98:
closes #484

### DIFF
--- a/src/javaMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerNameResolver.kt
+++ b/src/javaMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerNameResolver.kt
@@ -7,14 +7,19 @@ internal actual object KLoggerNameResolver {
 
   /** get class name for function by the package of the function */
   internal actual fun name(func: () -> Unit): String {
-    val name = func.javaClass.name
-    val slicedName =
-      when {
-        name.contains("Kt$") -> name.substringBefore("Kt$")
-        name.contains("$") -> name.substringBefore("$")
-        else -> name
+    return func::class.java.name.toCleanClassName()
+  }
+
+  private val classNameEndings = listOf("Kt$", "$")
+
+  private fun String.toCleanClassName(): String {
+    classNameEndings.forEach { ending ->
+      val indexOfEnding = this.indexOf(ending)
+      if (indexOfEnding != -1) {
+        return this.substring(0, indexOfEnding)
       }
-    return slicedName
+    }
+    return this
   }
 
   /** get class name for java class (that usually represents kotlin class) */
--- a/src/javaMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerNameResolver.kt
+++ b/src/javaMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerNameResolver.kt
@@ -1,13 +1,17 @@
 package io.github.oshai.kotlinlogging.internal
 
-import java.lang.reflect.Modifier
+import kotlin.reflect.KClass
 
 /** Resolves name of java classes */
 internal actual object KLoggerNameResolver {
 
   /** get class name for function by the package of the function */
   internal actual fun name(func: () -> Unit): String {
-    return func::class.java.name.toCleanClassName()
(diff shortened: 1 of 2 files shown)


oshai:
I am not sure what is the optimization here? Also, I think if $ or Kt$ appears twice it will yield different results than before.

scrat98:
@oshai it's unnecessary to search the same delimiter twice
1. contains = indexOf(delimeter) >= 0
2. substringBefore = substring(0, indexOf(delimeter))

scrat98:
@oshai also please note, that there is a problem with a tests that're not covering `name(func: () -> Unit): String` at all. The tests cover `name(forClass: Class<T>)` only, that have completely another logic to obtain the logger name https://github.com/oshai/kotlin-logging/issues/484#issuecomment-2704817334

oshai:
> @oshai it's unnecessary to search the same delimiter twice
> 
> 1. contains = indexOf(delimeter) >= 0
> 2. substringBefore = substring(0, indexOf(delimeter))

got it.

I am not sure wdym by tests not covering func. Anyway, it will be nice if you'll add unit test to the changed class.

scrat98:
@oshai I'll update the tests, but they will rely on **current** behavior. For example, tests expected to support inner classes

```kotlin
Arguments.of("io.github.oshai.kotlinlogging.internal.Foo\$Bar", Foo.Bar::class.java)
```

But currently Inner classes are not supported already, therefore it may seems that my changes break behavior, but actually not, this behavior was already changed. 
```kotlin
class Base {
  class Inner {
    private companion object {
      private val logger = KotlinLogging.logger { }
    }

    init {
        logger.info { "Hello" }
    }
  }
}

fun main() {
  Base.Inner()
}
```

```
18:28:14.314 [main] INFO  io.github.oshai.kotlinlogging.internal.Base - Hello
```

Please, see the comment here: https://github.com/oshai/kotlin-logging/issues/484#issuecomment-2704817334

scrat98:
@oshai please have a look

src/javaMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerNameResolver.kt
```diff
@@ -1,41 +1,28 @@
...
-      when {
-        name.contains("Kt$") -> name.substringBefore("Kt$")
-        name.contains("$") -> name.substringBefore("$")
-        else -> name
-      }
-    return slicedName
+    return name(func::class)
+  }
+
+  internal fun name(clazz: KClass<*>): String {
```
oshai, line 13: I think this method should be private

scrat98: @oshai historically this method was used in the tests, therefore it's not possible to make it private without rewriting tests structure. https://github.com/oshai/kotlin-logging/blob/master/src/javaMain/kotlin/io/github/oshai/kotlinlogging/internal/KLoggerNameResolver.kt#L21

I would prefer to keep it internal for now and if the changes from https://github.com/oshai/kotlin-logging/pull/481 will be accepted, then signature of the method will be changed from
```kotlin
internal actual fun name(func: () -> Unit): String
```

to
```kotlin
internal actual fun name(ref: Any): String {
```

and then we can update the tests to use instances of the classes, rather then KClass

oshai:
ok. checks are now failing - I think on code formatting.

scrat98:
Applied formatting, I'm sorry, completely forgot about it

scrat98:
Also I think it's worth to mention somewhere in the documentation that loggers in the inner classes always will have the name of the parent(base) class. Example from here https://github.com/oshai/kotlin-logging/pull/485#issuecomment-2781526598

Or create another issue if it supposed to be fixed

scrat98:
Created issue about logger name for inner class https://github.com/oshai/kotlin-logging/issues/495

PR #541, opened 2025-08-01
upgrade dokka to v2

oshai:
there are instructions here: https://kotlinlang.org/docs/dokka-migration.html


oshai:
this was reverted and now tracked in https://github.com/oshai/kotlin-logging/pull/543

PR #595, opened 2026-01-13
Remove CD workflow and add image rendering

jasminehaskins12:
Removed old CD workflow for GitHub releases and added image rendering code.


oshai:
Hi can you please explain this change?

PR #602, opened 2026-05-04
Fix #598: Allow external configuration of logStartupMessage via Properties and Env Vars

### LINKED ISSUE #598: Easier configuration of KotlinLoggingConfiguration.logStartupMessage
Per #587, setting `KotlinLoggingConfiguration.logStartupMessage` allows for not printing out the startup message, as long as it's called prior to any KotlinLogging initialization.  Managing that initialization is a little awkward, as at least in my case I've got a bunch of top level `private val log...`s  that get initialized before any call to main().  

I've been able to get it to work with a strategically placed `run{}` block like https://github.com/josephlbarnett/leakycauldron/commit/0622c91249c800a0fe319bd2dadac24e27edf8ff to take care of my main(), but the message shows up still in tests / other entry points, could there be an easier way to get this set throughout a given project?

oshai:
Addresses #598 by adding support for configuring KotlinLoggingConfiguration.logStartupMessage externally via System Properties and Environment Variables across all supported Kotlin Multiplatform platforms.

Currently, disabling the startup message programmatically can be awkward or impossible if top-level loggers initialize KotlinLogging during class loading before main() or test setups have a chance to run. Providing external configuration flags completely bypasses this initialization order constraint.

To maintain backward compatibility and avoid confusing users who rely on startup diagnostics, the default behavior remains true (opt-out).

Cross-Platform Implementation Details
The default value for logStartupMessage is now dynamically resolved during initialization based on the platform:

JVM & Android: Checks the JVM System Property kotlin-logging.logStartupMessage first, then falls back to the Environment Variable KOTLIN_LOGGING_STARTUP_MESSAGE.
Native & Darwin: Uses POSIX getenv to check the KOTLIN_LOGGING_STARTUP_MESSAGE environment variable.
JS & Wasm JS: Safely evaluates the Browser environment (window.KOTLIN_LOGGING_STARTUP_MESSAGE) or Node.js environment (process.env.KOTLIN_LOGGING_STARTUP_MESSAGE). Returns false if either resolves to "false" or boolean false.

Fixes #598

### DIFF
--- a/src/androidMain/kotlin/io/github/oshai/kotlinlogging/KotlinLoggingConfiguration.kt
+++ b/src/androidMain/kotlin/io/github/oshai/kotlinlogging/KotlinLoggingConfiguration.kt
@@ -36,7 +36,10 @@ public actual object KotlinLoggingConfiguration {
 
   @Volatile public actual var loggerFactory: KLoggerFactory = detectLogger()
 
-  public actual var logStartupMessage: Boolean = true
+  public actual var logStartupMessage: Boolean =
+    System.getProperty("kotlin-logging.logStartupMessage")?.toBoolean()
+      ?: System.getenv("KOTLIN_LOGGING_STARTUP_MESSAGE")?.toBoolean()
+      ?: true
 
   private fun detectLogger(): KLoggerFactory {
     if (System.getProperty("kotlin-logging-to-android-native") != null) {
--- a/src/darwinMain/kotlin/io/github/oshai/kotlinlogging/KotlinLoggingConfiguration.kt
+++ b/src/darwinMain/kotlin/io/github/oshai/kotlinlogging/KotlinLoggingConfiguration.kt
@@ -1,13 +1,22 @@
 package io.github.oshai.kotlinlogging
 
 import kotlin.concurrent.AtomicReference
+import platform.posix.getenv
+import kotlinx.cinterop.toKString
+import kotlinx.cinterop.ExperimentalForeignApi
+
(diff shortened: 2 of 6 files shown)


mikehearn:
Looks good, but I would also vote for just deleting it. There have been logging libraries for decades and I never saw one with a startup message like this. It seems very unlikely anyone would be confused and no tools should be relying on it for backwards compatibility.

Zordid:
When is this going to be released? I want to finally get this message disappear. I don't even know why you added this and defaulted it to appear... :-(

Zordid:
> Looks good, but I would also vote for just deleting it. There have been logging libraries for decades and I never saw one with a startup message like this. It seems very unlikely anyone would be confused and no tools should be relying on it for backwards compatibility.

Absolutely. It was added, but it hurts more than it adds. Everybody fights to get it off. Why was it added, what was the reasoning? When it was added lately, why wasn't backwards compatibility an issue? Why is it now for something so new, nobody can rely on it?

oshai:
The log startup message was introduced with the following design rationale:

Prevent User Confusion & Assist Debugging: With the shift to a unified multiplatform Universal Provider Architecture, the active logging provider (e.g., SLF4J, OSLog, Direct) became dynamically configurable and could vary by platform. Printing the active loggerFactory helps developers immediately confirm their exact runtime setup.

This was introduced in #584. 

I am working on finishing this PR and releasing it.

oshai:
released in 8.0.03 (will be available soon).

